PR #22, opened 2020-07-14
Use click outside

adrienpoly:
add a new behavior to track clicks outside an element

```js
//modal_controller.js
import { useClickOutside } from 'stimulus-use'

  connect() {
    useClickOutside(this)
  }

  clickOutside(event) {
    event.preventDefault()
    this.modal.close()
  }
}
```


fractaledmind:
There are some edge-cases discussed in a [StackOverflow answer](https://stackoverflow.com/a/3028037/2884386) (scroll thru the entire answer, as it has edits over time) which you may or may not wish to consider or deal with (I offer them simply because I have been thinking about this problem for a day or so myself):

1. is the target element currently visible
2. do we want to allow the user to click-and-drag inside the element, then release the mouse outside the element, without closing the element

However, my main point is that this looks awesome and this library keeps getting better and better. Thanks for the excellent work 😃

adrienpoly:
following to this conversation https://discourse.stimulusjs.org/t/best-practices-for-handling-clicks-outside-element/1266/4?u=adrienpoly

I have added a new event `click:outside` that is emitted so that we can also generic Stimulus actions

```html
<div data-controller="modal" data-action="click:outside->modal#close">
```

adrienpoly:
@fractaledmind do you have some example where `is the target element currently visible` is useful? 
I am having hard time finding examples.

About the rest of the SO thread they are some good topic on accessibility but it has more to do with closing a modal whereas here we just provide on particular tool that may be used to close a modal or do something else

Wondering if the blur event (pressing tab) should be considered as a click outside. Maybe an option on this could be good and it is activated by default

fractaledmind:
@adrienpoly: My first thought is something like a user/account dropdown in a header. You open the dropdown, you scroll the header out of the viewport, you scroll the dropdown out of the viewport, then you click. With the dropdown now no longer visible in the viewport, having that click close it would be an invisible effect. I don't think that is an awful UX, but it might be confusing. In general, it does make some sense to me to only have the `click:outside` event effect visible elements, but I don't think that is essential

adrienpoly:
Thanks @fractaledmind this is a very valid point. I started to introduce some options. I will add a visible option with a default to true

fractaledmind:
The new options look awesome @adrienpoly!

adrienpoly:
Thanks @fractaledmind I agree that the API is starting to look nice overall. Your feedback really helped

marcoroth:
Hey @adrienpoly and @fractaledmind, I really like how the `useClickOutside` module and it's API evolved. I think it's really elegant and I can see this module become very useful!

One more general thought: Would it make sense to prefix all the stimulus-use events? I don't know if this is really an issue, but I could imagine that some more general event name like `appear` could also being emitted by other common libraries. But it also makes the markup more verbose.

All in all I think this looks very solid and promising! Good job on this!

adrienpoly:
> One more general thought: Would it make sense to prefix all the stimulus-use events? I don't know if this is really an issue, but I could imagine that some more general event name like appear could also being emitted by other common libraries. But it also makes the markup more verbose.

I was having exactly the same though today. My idea was to keep by default the events without prefix but add a new option to add a custom prefix.

```javascript
const options = {
  eventPrefix: "su"
}
```

The emitted events are then `su-appear`, `su-disappear` 

I thought it would make more sens to have the prefix as an option in rather than always but open to other point of views

adrienpoly:
merging this one now some fine-tuning will be done in #33 with the event prefix

PR #27, opened 2020-07-24
Add useIdle

### LINKED ISSUE #2: useIdle
useIdle adds a new behaviour `idle` to trigger a callback when the user is idle on a page

**Inspiration:**

https://github.com/streamich/react-use/blob/master/docs/useIdle.md

marcoroth:
Hey @adrienpoly,

Here we go! This PR adds `useIdle` and resolves #2.

Before I add some tests for `useIdle` I wanted to discuss some things first:

### Amount of callbacks

I decided to add two callbacks, one that is getting triggered when the user is idle and one which is getting triggered as soon as the user is active again. In theory this could also be solved with one callback like this:

```js
idle(isIdle) {
  // isIdle would be true or false
}
```
But I thought it's somehow more intuitive to introduce two callbacks and to somehow match the behavior of  `useIntersection`.  Plus if you want to access the value of `isIdle` in your controller you can always just call `this.isIdle`.

### Callback names

I'm not 100% happy with the callback names `idle` and `returned`. Which ones would you prefer?

These are the ideas I had:

**`idle`**: 
* `away`
* `afk`
* `inactive`
* `absent`

**`returned`**: 
* `appear` (but this would interfere with `useIntersection`)
* `back`
* `present`
* `active`

### Options

Currently you can pass the options like this:

```js
useIdle(this, 3000, true, ['click'])
```

I noticed that you used an `options` object in `useIntersection` to pass the options.

But I thought that in probably 90% of the cases you just want to adjust the idle time. That's why I think it's more elegant to pass the options as separate parameters like this:

```js
useIdle(this, 5000)
```

----

I'm happy to discuss so we can iterate together! 😊


adrienpoly:
Excellent lots of good questions! 

Exactly what was needed at this point to finalise a bit the main structure of the API. 

Thanks 🙏

adrienpoly:
Here are a few comments. 
At this point, I will focus mostly on the API/ naming questions you had before getting into the code.

## Amount of callbacks

My initial tough was to go for a single callback with a boolean parameter but you are right that this does not match the `useIntersection` and I agree about having two.

## Callback names

looking at some other similar project they mostly use combo
https://firmfirm.github.io/f-idlejs/components/f-idlejs/ --> away / back
https://github.com/soixantecircuits/idle-js -->  idle/active

I think we should avoid `active` as it is too generic and a function name that someone could add in a typical controller.

About your proposal `idle/returned`. To be consistent with other functions (connect, appear, resize ...) we should keep infinitive verbs so `return` rather than `returned`

Personally I like the `away/back` it talks more to me than `idle/return` but I am really open for discussion on this

## Options

While I agree that the most common use case is just to set the timeout option, I would like that we keep the possibility to pass an object for the options and to standardize on this for all `useModule`.

I have a few ideas of additional params that we could have for all modules in the future.... 

**Examples :**
-  pass a custom `onIdle`  or `onReturn` function (you don't like `idle` or `away` you can pass your custom call back function)
-  In #22 I started to introduce some events emitted by those modules. If those events are generalized to all modules it would be best to have a boolean (dispatchEvent ??) to disable them if someone wants to avoid having those events dispatched with in the DOM.

One alternative would be to have a dual param type with an internal test to accept both

```js
useIdle(this, 2000)
useIdle(this, { initialState: false, timeout: 2000 })
```

marcoroth:
### Amount of callbacks

Great, then I will leave it as it is with the two callbacks.

### Callback names

After thinking a little bit about it I think `away/back` is a good option. First I had the feeling that we should keep `idle` because it's called `useIdle()`. But I think `away/back` is fine nevertheless as it describes the best what's actually happening.

=> I will update the callback names to `away/back` 👍 

### Options

> pass a custom `onIdle` or `onReturn` function (you don't like idle or away you can pass your custom call back function)

I can see this useful. Maybe it would also make sense to provide a way to rename the default names of the callback functions (could also be useful if would have a lot of callback names conflicts in the future). 

> One alternative would be to have a dual param type with an internal test to accept both

I kinda like this solution. I'd vote for this one. But we should make sure that we are consistent on how we handle options over all the modules. 

But maybe it's also worth to discuss if it makes sense to provide the options `initialState` and `events` in the first place.

adrienpoly:
Thanks @marcoroth for the update, I think it looks good.

For the options, having a way to configure the `initialState` and the `events` is a nice to have, and we should keep it.

Also if you have a bit of time and want to review #22 that would be great. 

You will see the event that is being introduced and the `withEvent` option. 
I think I will generalize this to all modules so that we could either have most of the logic inside the controller.

```js
appear() {
 this.fadeIn()
}
```

or to have the logic in the HTML markup

```html
<div data-controller="example" data-action="appear->example#fadeIn"></div>
```

src/use-idle/use-idle.ts
```diff
@@ -0,0 +1,73 @@
...
+    },
+    unObserve() {
+      events.forEach(event => {
+        window.removeEventListener(event, onEvent);
+      });
+      document.removeEventListener('visibilitychange', onVisibility);
+    },
+    disconnect() {
+      this.unObserve();
+      controller.disconnect();
```
adrienpoly, line 68: I had to do something like this to avoid having an infinite loop on the disconnect function

```js
  const controllerDisconnect = controller.disconnect

  Object.assign(controller, {
    disconnect() {
      this.unObserve()
      controllerDisconnect()
    },
  })
```

src/use-idle/use-idle.ts
```diff
@@ -0,0 +1,73 @@
...
+const oneMinute = 60e3;
+
+export const useIdle = (controller: IdleController, ms: number = oneMinute, initialState: boolean = false, events: string[] = defaultEvents) => {
+  let isIdle = initialState;
+  let timeout = setTimeout(() => {
+    isIdle = true
+    dispatchAway();
+  }, ms);
+
+  const method = (methodName: string): Function => {
```
adrienpoly, line 13: In case this branch gets merged after #22, I have extracted this function into a helper function share between all modules. But don't worry about this for now

src/use-idle/use-idle.ts
marcoroth: Ahh, that makes sense. You are "saving" the old `disconnect` function before reassigning the new one with `Object.assign` to later call the old one again. I will update it 😊

src/use-idle/use-idle.ts
marcoroth: Yeah I've seen that, makes sense 👍

README.md
```diff
@@ -44,6 +44,7 @@ yarn add stimulus-use
   |-----------------------|-------------|---------------------|
   | [`useIntersection`](./docs/use-intersection.md) </br> `IntersectionController` | Tracks the element's intersection and adds **appear**, **disappear** callbacks to your controller.|`appear`</br> `disappear`|
   |[`useResize`](./docs/use-resize.md) </br>`ResizeController`|Tracks the element's size and adds a new lifecyle callback **resized**.|`resized`|
+  |[`useIdle`](./docs/use-idle.md) </br>`IdleController`| Tracks if the user is idle on your page and adds **away** and **back** callbacks to your controller.|`away`<br/>`back`|
```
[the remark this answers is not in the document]
marcoroth: Ooh there is a `docs/home.md`, I havn't seen that file 😅 I can update it myself for now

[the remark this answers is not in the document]
adrienpoly: yeah it is not obvious and not really DRY... will need to find a better startegy ultimately

marcoroth:
Thanks for the review @adrienpoly! 
I addressed your feedback by reworking the options and the addition of the two new `away` and `back` events.

I'm just not sure if it makes sense to dispatch those events on `controller.element` or if I would make more sense to dispatch them on `window` or `document`. The thought behind this is that you can't have different `isIdle` states across multiple controllers so you could also emit that event once "globally".

adrienpoly:
> I'm just not sure if it makes sense to dispatch those events on `controller.element` or if I would make more sense to dispatch them on `window` or `document`. The thought behind this is that you can't have different `isIdle` states across multiple controllers so you could also emit that event once "globally".

The events have by default a bubble true option. My idea was to emit them from the controller element and if needed we can always listen to them with `appear@window->animation#fadeIn`

I have seen that in order to use the function `extendedEvent` you created some dummy `back` and `away` events. I started to work on adding appear, resize etc events and those also are not a result of a real event at first as was click:outside. I have changed the params of `extendedEvent` in #33 the event is now optional.

Sorry that will make a few change on your side....

adrienpoly:
@marcoroth if it is ok for you I'll merge it as it is right now.
Once I get #34 merged I will add a few basic test to it

marcoroth:
@adrienpoly if you would like to merge it go ahead. 

I was working on tests but couldn't find a nice way to handle the "waiting in real-time" to simulate that a user is idle. I need to have another look at it to see how I can get that solved.

adrienpoly:
@all-contributors please add @marcoroth for code

PR #49, opened 2020-09-15
introduce useVisibility

adrienpoly:
example for playing a video only when the tab is active

```js
//video_controller
  connect() {
    useVisible(this)
    this.player = new VideoPlayer()
  }

  visible() {
    this.player.play()
  }

  invisible() {
    this.player.pause()
  }
}
```


docs/use-visibility.md
```diff
@@ -0,0 +1,101 @@
+# useClickOutside
```
marcoroth, line 1: I think this should be `useVisibility`

docs/use-visibility.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+**controller** : a Stimulus Controller (usually `'this'`)
+
+**options** :
+
+| Option| Description |&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;Default value&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;|
+|-----------------------|-------------|---------------------|
+| `dispatchEvent` | Whether to dispatch a `click:outside` event or not.| `true` |
+|`eventPrefix`| Whether to prefix or not the emitted event. Can be a **boolean** or a **string**.<br>- **true** prefix the event with the controller identifier `card:click:outside` <br>- **someString** prefix the event with the given string `someString:click:outside` <br>- **false** to remove prefix  |true|
```
marcoroth, line 20: Maybe we can rewrite this in the context of `useVisibility`

marcoroth (commented):
This looks very good @adrienpoly! 
I just noticed two small things which could be easily improved, but other than that this is very solid! 👏🏼

adrienpoly:
Thanks for your review @marcoroth 

Will need also to check a bit the cross-browser compatibility but will do that when I get rid of a few flaky tests and get SauceLabs at full speed  👉 #48 to have a similar coverage than Stimulus.

Also I tried to package the mixin within a [Class](https://github.com/stimulus-use/stimulus-use/blob/master/src/use-visibility/use-visibility.ts#L14) `UseVisibility` rather than a pure function. I think it could help to make the code more readable. I will play a bit with it first before converting other mixins to it.

abozhinov:
Hi @adrienpoly, 
I try to use useVisibility but every time I swipe my desktop screens on mac the visible function is activated. Maybe I didn't understand the purpose of this mixin but I think it's useful to know the state of the element.

PR #50, opened 2020-10-06
Introduce useWindowSize

### LINKED ISSUE #1: useWindow Resize
Add new behavior for tracking windowResize event similar to `useResize`

inspiration :

https://github.com/streamich/react-use/blob/master/docs/useWindowSize.md

marcoroth:
This PR introduces the `useWindowSize` behaviour.

Points to discuss: 
* [x] ~Does it make sense to reuse the `resize` lifecycle name? I think it's fine.~ => renamed to `windowResize`
* [x] I left out the `dispatchEvent` in this behavior as it doesn't really add any value
* [x] This behavior also doesn't have any options at all. Seems like there isn't the need for it
* [ ] A way to test behaviors like this

This PR Resolves #1.


adrienpoly:
Thanks @marcoroth this looks very good. I am ok with your comments for the event and options. 

As you said this one is a basic behavior on the top of a single standard event. It doesn't add as much value as others observers that are not pure JS events. Still it provides a nice and clean API to get the width and height of the window in a global way without explicitly declaring an action `resize@window->....`

While I like having explicit html I feel like sometime having a strict behavior for a controller independent of the html tags is nice.

I just need to think a bit about the resize naming my initial though was `windowResize` to be more explicit and without any overlap with the resize observer. But I agree that there shouldn't be any overlap.

Let me think a bit about this one.

This being said (not for V1) I was thinking about an option to supply a custom callback function. Something like 
```js
useWindowResize(this, { onResize: this.customHandleResize })
``` 
So if we have the option then there is always a solution to manage overlap for the callback naming in the unlikely case where both useResize and useWindowResize would be used and this would vote for keeping a simple `resize`name....

Last I was looking at testing behavior like this (when working on visibility). I don't think we can test it with the current setup but this would require a more E2E testing solution such a Cypress. Do you have any thoughts/experience on this?

adrienpoly:
Hey after some though over the week-end I think it is best to keep it explicit as `windowResize` rather than `resize`

marcoroth:
Hey @adrienpoly. Sure, I will change the callback name to `windowResize`. 

And re: Cypress - I've used it in the past. I try to come up with a good setup/solution to test behaviors like this.

marcoroth:
Alright, I've updated the callback name and introduced cypress with some specs for the `windowSize` behavior.  
I'm happy to add more specs for the other behaviors once we decided how we are going to test functionality like this.

I don't know where you want to run the E2E tests, in CircleCI or in GitHub Actions?
Here are setup instructions for both:
https://docs.cypress.io/guides/guides/continuous-integration.html#CircleCI
https://github.com/cypress-io/github-action#headless

It's hard to test it in CircleCI because it's not activated on PRs, maybe you can take a short look how to set that up.

adrienpoly:
This looks great @marcoroth I'll merge it and try to include it in CircleCI

marcoroth:
Cool, let me know if you need help!

PR #51, opened 2020-10-20
Memoize

julianrubisch:
This is a proof of concept PR and lacks tests, proper typing and documentation.

To test, `yarn build` and start the playground, observe the console.

Getters `a` and `b`, which are stored in the `static memoizedGetters` array, are first memoized, then they don't fire anymore.

Getter `c` is not memoized and gets called every time.


src/use-memo/use-memo.ts
```diff
@@ -0,0 +1,14 @@
+// import { MemoController } from './memo-controller'
+// import { method } from '../support'
+
+const memoize = (object: object, name: string, value: any) => {
+  console.log(`memoizing ${name} with ${value}`)
```
julianrubisch, line 5: this `console.log` will of course be removed, only for visualization purposes.

adrienpoly (changes requested):
Thanks @julianrubisch this is an excellent contribution

I am ok with getting a first version like that focusing on getters

## TS typing
I made a few comments and tried to improve a bit the TS typing. @marcoroth if you have some suggestion 

## Naming

I find `memoizedGetters` a bit long and hard to remember.
My suggestion would be to simply use `memo` but I understand that we lose the idea that it is for getters only.

If we think that tomorrow this mixin could memoize also functions in addition to getters then I think it would be best to go directly with `memo` to avoid some breaking change/complexity down the road. 

@marcoroth if you have some naming suggestion 

## Documentation 
For the documentation we will need to create a new category for this type of mixin

`React-Use` has the following
- sensors
- ui 
- animation
- side-effects
- lifecylces
- state

`createMemo` is within the state category

But I am not sure having a `state` category for Stimulus really makes sens?

Currently, there is the `Observers` category and a stand alone `ApplicationController`. 
`useDispatch` will be extracted from the application controller to its own mixin. Maybe with in a group of `communication` related mixins?

Suggestion for `useMemo` 
- Helpers
- Performances

src/use-memo/use-memo.ts
```diff
@@ -0,0 +1,14 @@
...
+  Object.defineProperty(object, name, { value })
+  return value
+}
+
+export const useMemo = (controller: any) => {
+  controller.constructor.memoizedGetters.forEach((getter: string) => {
+    memoize(controller, getter, controller[getter])
+  })
+}
```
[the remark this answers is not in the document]
julianrubisch: Ah... (controller.constructor as any) was tripping me up, thanks!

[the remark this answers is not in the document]
adrienpoly: yeah I know I struggled a few time on this one before. They might be a stronger type solution but it makes the job right now

julianrubisch:
Updated the typings.

Re documentation: What about an **Optimization** category?

adrienpoly:
I like Optimization 👍

julianrubisch:
Superb, I'm going to write up some docs later 👍

marcoroth:
I really like this addition @julianrubisch! Pretty straight-forward and concise solution!

@adrienpoly about the naming: I agree with you. If we want to make it future-proof  `memo` would fit nicely. I think I would name it `memos` (in plural) because there can be multiple memoized getters and/or functions in the future. Plus it would match the Stimulus `targets` naming which is also plural.

julianrubisch:
Should we worry about invalidating later?

src/use-memo/use-memo.ts
```diff
@@ -0,0 +1,12 @@
+import { Controller } from 'stimulus'
+
+const memoize = (controller: Controller, name: string, value: any) => {
+  Object.defineProperty(controller, name, { value })
+  return value
+}
+
+export const useMemo = (controller: Controller) => {
+  ;(controller.constructor as any).memos?.forEach((getter: string) => {
+    memoize(controller, getter, Reflect.get(controller, getter))
```
adrienpoly, line 10: ok looks like Reflect is not supported on IE11
My initial plan was to try to get the same level of browser compatibility as Stimulus. I got a free open source saucelab account that I still need to finish setting up
So either we add a polyfill to the documentation 
https://github.com/tvcutsem/harmony-reflect (and to our test environment)
or 
do something like that `(controller as any)[getter]`

solution B looks easier

playground/controllers/memo_controller.js
```diff
@@ -0,0 +1,33 @@
...
+  get b() {
+    console.log('getting b')
+    return 2
+  }
+
+  get c() {
+    console.log('getting c')
+    return 3
+  }
+}
```
adrienpoly, line 33: I think it would be nice to provide a more realistic example

suggestion 
```js
import { useMemo } from 'stimulus-use'
import { intensiveNormalizeTask } from 'some-module'

  static memos = ['normalizedTex']
  static targets = ['longText']

  connect() {
    useMemo(this)

    this.normalizedText // process the text and cache the result
    this.normalizedText // gets the results from the cache
  }

  get normalizedText() {
    return intensiveNormalizeTask(this.longTextTarget.textContent)
  }
}
```

adrienpoly:
@julianrubisch this real example is really cool.
I realized that I commented the wrong file my initial idea was to change the example in the doc.
don't worry I'll take care of that
🙏 for this excellent contribution

adrienpoly:
@all-contributtors add @julianrubisch  to code

adrienpoly:
@all-contributors add @julianrubisch to code

PR #53, opened 2020-10-23
introduce useDebounce

adrienpoly:
Introducing a new mixin for debouncing controllers functions. it follows a similar pattern as [useMemo](https://github.com/stimulus-use/stimulus-use/blob/master/docs/use-memo.md) with a static array "debounces" of function names to debounce

```js
import { ApplicationController, useDebounce } from 'stimulus-use'

export default class extends ApplicationController {
  static debounces = ['add']

  connect() {
    useDebounce(this, {delay: 100})
  }

  add() { 
  // this function is debounced
 }
}
```


adrienpoly:
👋  @julianrubisch @marcoroth WDYT?

julianrubisch:
good idea! we should look if there's some overlap with https://github.com/hopsoft/debounced

julianrubisch:
oh it's already a PR, I thought it's a feature request... I'll take a look then

julianrubisch:
looks all good to me!

marcoroth:
Looks good to me as well! The only thing is asked myself: Should you be able to specify different delays for the different controller functions? 
But I think it's fine that you can just set it globally - also because I can't think of nice and intuitive syntax for that.

julianrubisch:
we could go ahead and create `useThrottle` as well 😎

adrienpoly:
That was my plan but wanted to play a bit first with the Debounce api

adrienpoly:
@all-contributors add @julianrubisch to review

adrienpoly:
@all-contributors add @marcoroth to review

PR #69, opened 2020-11-16
Update application-controller.md

woto:


woto:
Hmmm, but: https://github.com/stimulus-use/stimulus-use/blob/ed93f68faa036c8ef6a256abc2c9453642a57226/spec/use-application_spec.js#L16
Sorry. Don't have time to figure out the problem. But it does not work as it is written in the documentation. May be it will help.

adrienpoly:
This is a breaking change that was introduced with version 0.17 and part of the extraction of useDispatch to its own mixin 

https://github.com/stimulus-use/stimulus-use/releases/tag/v0.17.0

Unless you see an error with 0.17+ I think we can close this PR

woto:
Ok, I'll try to update later. Thanks for the explanation and the library 😊. Will reopen if it will not.

PR #72, opened 2020-11-25
docs: fix useWindowResize link in README

dancallaghan:


adrienpoly:
Thanks for the fix

@all-contributors add @dancallaghan to doc

PR #75, opened 2020-12-02
Was trolling through and found this typo.

ParamagicDev:


adrienpoly:
Thanks a lot!! 
I will double check but it probably means that this settings is not required as it was working like that 🤔

adrienpoly:
@all-contributors add @ParamagicDev to bug

PR #77, opened 2020-12-03
introduce useTransition

adrienpoly:
New mixin to handle animations and transition ala Vue/Alpine

## Simple Tailwind Dropdown example

```js
import { useClickOutside, useTransition } from 'stimulus-use'

  static targets = ['content']

  connect() {
    useClickOutside(this)
    useTransition(this, { element: this.contentTarget })
  }

  clickOutside(e) {
    this.hide()
  }
}
```

```html
<div class="relative" data-controller="dropdown">
  <div>
    <button
      data-action="click->dropdown#toggle"
      type="button"
      class="inline-flex items-center px-2.5 py-1.5 border border-transparent text-xs font-medium rounded text-indigo-700 bg-indigo-100 hover:bg-indigo-200 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500">
      Dropdown
    </button>
  </div>
  <div class="origin-top-right absolute left-0 mt-2 w-48 rounded-md shadow-lg py-1 bg-white ring-1 ring-black ring-opacity-5"
  data-target="dropdown.content"
  data-transition-enter="transition ease-out duration-300"
  data-transition-enter-active="transform opacity-0 scale-95"
  data-transition-enter-to="transform opacity-100 scale-100"
  data-transition-leave="transition ease-in duration-300"
  data-transition-leave-active="transform opacity-100 scale-100"
  data-transition-leave-to="transform opacity-0 scale-95"
  role="menu"
  aria-orientation="vertical"
  aria-labelledby="user-menu">
    <a href="#" class="block px-4 py-2 text-sm text-gray-700 hover:bg-gray-100" role="menuitem">Your Profile</a>
    <a href="#" class="block px-4 py-2 text-sm text-gray-700 hover:bg-gray-100" role="menuitem">Settings</a>
    <a href="#" class="block px-4 py-2 text-sm text-gray-700 hover:bg-gray-100" role="menuitem">Sign out</a>
  </div>
</div>
```

## Alpine or Vue syntax

Alpine classes are enterStart/enterEnd whereas Vue is enterActive/enterTo
This mixin supports both syntax to make it easier to convert Tailwind codes as an example

#### VueJS

```html
  data-transition-enter="transition ease-out duration-300"
  data-transition-enter-active="transform opacity-0 scale-95"
  data-transition-enter-to="transform opacity-100 scale-100"
  data-transition-leave="transition ease-in duration-300"
  data-transition-leave-active="transform opacity-100 scale-100"
  data-transition-leave-to="transform opacity-0 scale-95"
```

#### AlpineJS

```html
  data-transition-enter-class="transition ease-out duration-300"
  data-transition-enter-start-class="transform opacity-0 scale-95"
  data-transition-enter-end-class="transform opacity-100 scale-100"
  data-transition-leave-class="transition ease-in duration-300"
  data-transition-leave-start-class="transform opacity-100 scale-100"
  data-transition-leave-end-class="transform opacity-0 scale-95"
```

- [ ] add tests
- [ ] fix edge cases (recover inital state)
- [ ] add documentation


adrienpoly:
updated the description of the PR click submit too soon 😄

julianrubisch:
Just saw that PR and stopped by to say feel free to look at https://github.com/excid3/tailwindcss-stimulus-components

I kind of solved the hard parts there already...

marcoroth:
I love the new playground! And `useTransition` looks promising!

julianrubisch:
I yearn for stimulus 2. valuechanged callbacks will really make that as easy as to switch an attribute on and off

adrienpoly:
@julianrubisch yes I saw your [PR](https://github.com/excid3/tailwindcss-stimulus-components/pull/44) 

I am not that familiar with this overall animation/transition echo system. 
I looked at both Vue and Alpine API as they seem to be the widest users as well as Tailwind UI documentation.

In Tailwind-Stimulus-Components the syntax of your PR is a bit different 
invisible/visible/entering/Enter... does it come from another framework ?
I like the naming.

@marcoroth thanks for the playground it was really required to pimp it a bit
next step is probably to make it a Bridgetown demo site on a GitHub page (poke @ParamagicDev  😄)

julianrubisch:
Basically I just used tailwindui‘s naming 😉

nfilzi:
@julianrubisch your wish has been heard 🎉 
https://github.com/stimulusjs/stimulus/pull/202

src/use-transition/use-transition.ts
```diff
@@ -0,0 +1,154 @@
...
+      show()
+    }
+  }
+
+  async function transition(element: Element, initialClass: string[], activeClass: string[], toClass: string[]) {
+    element.classList.add(...initialClass);
+    element.classList.add(...activeClass);
+    await nextFrame();
+
+    element.classList.remove(...activeClass);
```
jduff, line 100: Something I ran into when implementing transitions was there were a few cases where I had a class on the element initially, but it was also in the active transition. I started with a line exactly like this, but that meant when the transition finished the class that was initially on the element was removed as well. Alpine handles this by storing the initial classes and leaving those in place when removing the active classes https://github.com/alpinejs/alpine/blob/master/src/utils.js#L427

If you're interested the package I have been using is [here](https://github.com/mmccall10/el-transition) and the patch for this issue [here](https://github.com/mmccall10/el-transition/pull/2). Your approach looks very similar, looking forward to switching over to your version and dropping a dependency!

src/use-transition/use-transition.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function getAttribute(name: string, options: TransitionOptions, dataset: DOMStringMap) {
+  const datasetName = `transition${name[0].toUpperCase()}${name.substr(1)}`
+  const datasetAlpineName = (alpineNames as any)[name]
+  const classes = (options as any)[name] || dataset[datasetName] || dataset[datasetAlpineName] || ""
+  return classes.split(" ")
+}
+
+async function afterTransition(element: Element): Promise<Function> {
```
panda-madness, line 133: I've implemented something similar before. Here's my implementation for `afterTransition` that handles multiple different transition durations on an element:
```
function singleTimeValueToMs(duration: string): number {
    const matches = duration.match(/^(-?[0-9.]+)(s|ms)$/);

    const invalidValueError = new Error('Invalid time value');

    if (matches === null) {
        throw invalidValueError;
    }

    if (matches.length !== 3) {
        throw invalidValueError;
    }

    const [match, length, unit] = matches;

    if (unit === 's') {
        return Number(length) * 1000;
    }

    return Number(length);
}

function multiTimeValueToMs(durations: string): number[] {
    return durations.split(',')
        .map(str => str.trim())
        .map(singleTimeValueToMs);
}

function afterTransition(element) {
    return new Promise(resolve => {
        const durations = multiTimeValueToMs(
            window.getComputedStyle(element).transitionDuration
        );

        setTimeout(resolve, Math.max(...durations));
    });
}
```

src/use-transition/use-transition.ts
adrienpoly: Thanks @jduff I started to feel similar issues yesterday while testing it. 
I agree that this should be addressed. Will look a bit more into Alpine code to make sure we have something equivalent in terms of edges cases.

src/use-transition/use-transition.ts
adrienpoly: Thanks @panda-madness will definitely check it out

rainerborene:
Great work @adrienpoly. I have something similar on our codebase. I suggest renaming the controllers methods `show` and `hide` to `enter` and `leave` respectively because could be other transition scenarios beyond simple show and hide effects.

adrienpoly:
I have renamed the function to `enter`, `leave`and `toggleTransition` and the inner boolean to `transitioned` as per @rainerborene suggestions. 

I will open 2 issues to cover @panda-madness https://github.com/stimulus-use/stimulus-use/pull/77#discussion_r535858372 and @jduff https://github.com/stimulus-use/stimulus-use/pull/77#discussion_r535778033

in the meantime I will publish a beta (undocummented) version of this mixin for further testing.

rainerborene:
Just upgraded to `useTransition`. Thank you so much! :pray:

guillaumebriday:
Great!! 🥳 Thank you @adrienpoly I will update many of my https://github.com/stimulus-components/stimulus-components packages to use this feature!

adrienpoly:
@guillaumebriday this great, they are still quite a few edge cases as listed above, and I think I saw a new one while testing but needs to be confirmed. 

Don't hesitate to open an issue if you find something

PR #87, opened 2020-12-14
Avoid overriding defaultOptions directly

tsmd:
When calculating options using defaultOptions, the first argument of Object.assign should be {}.
Otherwise the defaultOptions itself would be rewritten.

Thanks for maintaining a great product!


adrienpoly:
🙏  Thanks very nice catch!

all-contributors please add @tsmd to bug

adrienpoly:
@all-contributors please add @tsmd to bug

adrienpoly:
patch available in [0.21.2](https://github.com/stimulus-use/stimulus-use/releases/tag/v0.21.2)

PR #96, opened 2021-01-10
New Observer - useHover

Sub-Xaero:
Hi there,

Love this project, great work on it! :+1 

I ended up implementing this extra feature for my own projects so I thought I'd contribute it back if you're interested.

I've tried to match your style and code conventions. Let me know if you want me to make any changes.

### DIFF
--- a/src/use-hover/use-hover.ts
+++ b/src/use-hover/use-hover.ts
@@ -1,40 +1,56 @@
-import { method } from '../support/index'
+import { StimulusUse, StimulusUseOptions } from '../stimulus_use'
+import { method } from '../support'
 import { HoverController } from './hover-controller'
 
-export interface HoverOptions {
+export interface HoverOptions extends StimulusUseOptions {
   element?: Element
 }
 
-export const useHover = (controller: HoverController, options: HoverOptions = {}) => {
-  const targetElement: Element = options?.element || controller.element
+export class UseHover extends StimulusUse {
+  controller: HoverController
+  targetElement: Element
 
-  const onEnter = () => {
-    controller.mouseEnter && method(controller, 'mouseEnter').call(controller)
-  }
-  const onLeave = () => {
-    controller.mouseLeave && method(controller, 'mouseLeave').call(controller)
-  }
-
-  const controllerDisconnect = controller.disconnect.bind(controller)
-
-  const observe = () => {
-    targetElement.addEventListener('mouseenter', onEnter)
-    targetElement.addEventListener('mouseleave', onLeave)
+  constructor(controller: HoverController, options: HoverOptions = {}) {
+    super(controller, options)
+    this.targetElement = options?.element || controller.element
+    this.controller = controller
+    this.enhanceController()
+    this.observe()
   }
 
-  const unobserve = () => {
-    targetElement.removeEventListener('mouseenter', onEnter)
-    targetElement.removeEventListener('mouseleave', onLeave)
+  observe = () => {
+    this.targetElement.addEventListener('mouseenter', this.onEnter)
+    this.targetElement.addEventListener('mouseleave', this.onLeave)
   }
 
-  Object.assign(controller, {
-    disconnect() {
-      unobserve()
+  unobserve = () => {
+    this.targetElement.removeEventListener('mouseenter', this.onEnter)
+    this.targetElement.removeEventListener('mouseleave', this.onLeave)
+  }
+
+  private onEnter = () => {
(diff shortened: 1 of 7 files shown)


src/use-hover/use-hover.ts
```diff
@@ -0,0 +1,40 @@
...
+    disconnect() {
+      unobserve()
+      controllerDisconnect()
+    }
+  })
+
+  observe()
+
+  return [observe, unobserve] as const
+}
```
adrienpoly, line 40: Thanks for trying to match the current code style. 

I have recently introduced a debug mode (#85) and as part of the debug mode start to refactor a bit the code to rely more on `Classes` rather than pure functions. There is a base class `StimulusUse` that provides a few common features (currently mostly the log function). 
I am in the process of converting the different mixins to this pattern, but it is still a WIP. Currently, `useDispatch` & `useVisibility` are based on this. It would be great if `useHover` could rely on this pattern out of the box.

I haven't tested it but here is a quick conversion of the code to this format

```js
import { method } from '../support/index'
import { StimulusUse, StimulusUseOptions } from '../stimulus_use'
import { HoverController } from './hover-controller'

export interface HoverOptions extends StimulusUseOptions {
  element?: Element
}

export class UseHover extends StimulusUse {
  controller: HoverController
  targetElement: Element

  constructor(controller: HoverController, options: HoverOptions = {}) {
    super(controller, options)
    this.targetElement = options?.element || controller.element
    this.controller = controller
    this.enhanceController()
    this.observe()
  }

  observe = () => {
    this.targetElement.addEventListener('mouseenter', this.onEnter)
    this.targetElement.addEventListener('mouseleave', this.onLeave)
  }

  unobserve = () => {
    this.targetElement.removeEventListener('mouseenter', this.onEnter)
    this.targetElement.removeEventListener('mouseleave', this.onLeave)
  }

  // private

  private onEnter = () => {
    this.controller.mouseEnter && method(this.controller, 'mouseEnter').call(this.controller)
    this.log("mouseEnter", { hover: true })
  }

  private onLeave = () => {
    this.controller.mouseLeave && method(this.controller, 'mouseLeave').call(this.controller)
    this.log("mouseLeave", { hover: false })
  }

  private enhanceController() {
    const controllerDisconnect = this.controller.disconnect.bind(this.controller)

    const disconnect = () => {
      this.unobserve()
      controllerDisconnect()
    }

    Object.assign(this.controller, { disconnect })
  }
}

export const useHover = (controller: HoverController, options: HoverOptions = {}) => {
  const observer = new UseHover(controller, options)
  return [observer.observe, observer.unobserve] as const
}
```

adrienpoly (changes requested):
Hey thanks this is an excellent addition that I had on my list so 👍. 

While some can object that this can easily be achieved by adding a `data-action="mouseenter->controller#enter` I fully support that in some case you just want to add a controller to a div and this controller comes fully featured, and you don't have to setup anything more.

I have made a few request change one that is mostly because I am transitioning the code to `classes` and also to support the debug feature (https://github.com/stimulus-use/stimulus-use/blob/master/docs/debug.md)

Let me know if you have any questions.

Also would be interested to get a few of your real world use cases. This is a section that I would like to had to the doc at some point

src/use-hover/use-hover.ts
Sub-Xaero: Ooh nice, seems to work in testing, pushed those changes :) 👍

Sub-Xaero:
Thanks for the review, I've made those changes 🙂 

Yeah, exactly as you say, the main use-case for writing it was for drop-in controllers where you can just specify the `data-controller="foo"` and the whole behaviour gets installed by the `connect()`, without having to do too much wiring in the DOM. That and having a consistent controller structure when composing multiple Stimulus-Use behaviours together, and the fire-and-forget convenience of having Stimulus-Use do the event handler attaching and cleanup for you.

My personal real-world use case is still in development with not much to show for it yet, but takes inspiration from Github's hover-cards, i.e. 
image
Being able to hover over a link and get a preview card to show - which could either involve using hover to invoke another JS framework, or do some AJAX loading.

adrienpoly:
Thanks this is looking good to me. I think there is a missing export in the main `index.ts` but I will add that before publishing a release.

@all-contributors please add @Sub-Xaero to code

adrienpoly:
@all-contributors please add @Sub-Xaero for code

Sub-Xaero:
@adrienpoly  That's awesome, thanks! Let me know if there are any other features off of your list you want me to have a poke at.

PR #98, opened 2021-01-13
Introduce useMutation

Sub-Xaero:
Sorry - Me again. Not sure if this was one that was on your list of desired features, but it's here if you want it. 

I was doing some work with the new @hotwired/turbo framework and needed to know when a turbo stream response mutated a particular section of the DOM, hence this controller. 

Happy to tweak the API as I wasn't sure whether it was better to pass the whole `entries` array to the enhanced controller's callback, or whether to pass them one by one. Chose to do that as it mirrors the behaviour of the underlying MutationObserver. 

Similarly, in the `log` logic, I'm logging the whole `entries` array of `MutationRecord`s that the observer gives to the callback, again - so that it matches the underlying Observer.


adrienpoly:
👍interesting I have a POC somewhere in a branch for this but never really had time to finish it. This is a great addition. I’ll look into your code and details a bit later

README.md
```diff
@@ -55,6 +55,7 @@ We got you covered 👉 [stimulus-use.github.io/stimulus-use](https://stimulus-u
   |[`useHover`](./docs/use-hover.md)|Tracks the user's mouse movements over an element and adds **mouseEnter** and **mouseLeave** callbacks to your controller.|`mouseEnter` `mouseLeave`|
   |[`useIdle`](./docs/use-idle.md)| Tracks if the user is idle on your page and adds **away** and **back** callbacks to your controller.|`away`</br> `back`|
   |[`useIntersection`](./docs/use-intersection.md) | Tracks the element's intersection and adds **appear**, **disappear** callbacks to your controller.|`appear`</br> `disappear`|
+  |[`useMutation`](./docs/use-mutation.md) | Tracks mutations on an element, its attributes and/or subtree. Adds a **mutated** callback to your controller.|`mutated`|
```
adrienpoly, line 58: other mixin are using infinitive tense for the name of the callback. 

`resize` rather than `resized`
`appear` rather than `appeared`

Can we rather use `mutate`

adrienpoly:
Other than the consideration for the callback naming this looks very solid 👏.

When I looked at it initially I wanted to add options to easily filter Stimulus Target to simplify the MutationObserver options. 
Something that could somehow solve this issue https://github.com/hotwired/stimulus/issues/336 reported by @leastbad 

But that as it is, is a GREAT first step and we can think further down the road how options/filters can be abstracted to common paterns in Stimulus

Sub-Xaero:
@adrienpoly Great, I've made those changes.

Yes, I've been wanting those target added and removed callbacks since forever as well. I'll see if I can knock something up. It's complicated enough logic that it probably deserves its own enhancement function, otherwise the useMutation one will get slightly unruly.

leastbad:
@Sub-Xaero You rock, sir!

adrienpoly:
I added some basic test to play a bit with it. One thing that I noticed is that we would get an error with the default options

`useMutation(this)` 

It is the described behavior of the mutation observers, as written in the doc. 

Now I was wondering if by your experience we could have defaults that would match typical use cases for Stimulus and prevent an error with default values?

With in the test there is a commented scenario for playing the defaults values

```js
const scenarios = [
  // {
  //   name: 'default options',
  //   fixture: fixtureBase,
  //   options: {},
  //   answers: {}
  // },
  {
    name: 'childList: true',
    fixture: fixtureBase,
    options: { childList: true },
    answers: {}
  }
]
```

Sub-Xaero:
Well, I think https://github.com/stimulus-use/stimulus-use/pull/99 meets the more common use case that people might have for this in knowing when targets come in and out. And the Stimulus 2.0 values API covers the average use case for watching data-attributes change. 

Which might mean that this is fine to be left open to be "configurable" and not make any assumptions about its use. Maybe throw our own more meaningful error than letting the MutationObserver throw. 
But, if we have to specify a default `childList` would be the most sensible to my mind, listening to `attributes` is going to fire a lot more often that dom insertions/removals I would have thought.

src/use-mutation/use-mutation.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.controller = controller
+    this.options = options
+    this.observer = new MutationObserver(this.mutation)
+
+    this.enhanceController()
+    this.observe()
+  }
+
+  observe = () => {
+    this.observer.observe(this.targetElement, this.options)
```
adrienpoly, line 31: maybe we could wrap it in a try/catch and report the error to Stimulus ErrorHandler

```suggestion
   try {
      this.observer.observe(this.targetElement, this.options)
    } catch (error) {
      this.controller.application.handleError(error,"At a minimum, one of childList, attributes, and/or characterData must be true",{})
    }
```

Sub-Xaero: Sounds entirely reasonable!

adrienpoly:
LGTM thanks for this great addition 🚀

adrienpoly:
@all-contributors please add @Sub-Xaero for doc and code

PR #99, opened 2021-01-14
Introduce useTargetMutation

Sub-Xaero:
As discussed in https://github.com/stimulus-use/stimulus-use/pull/98 , I've tried to knock together a nice mutation observer wrapper that generates `added` and `removed` callbacks for the specified targets. 

I would have liked to have just been able to handle the whole set of specified targets, but Stimulus wraps them in a `TargetSet` such that you can't get at the static array of strings anymore. At least not that I could find. 

But specifying the ones you care to watch, seems like a reasonable compromise. 

I'll need to change the names of the defined methods, to be `[target]TargetAdd`, to match the rest of the library, as before, but I knocked this up at a crazy hour of the morning and ran out of time. 

Let me know what you think.


adrienpoly:
Thanks @Sub-Xaero this is 🚀. 
I went through the code quickly, this is an amazing start. I spot a few ideas of improvements that I will summarize later in the day.

Sub-Xaero:
@adrienpoly Absolutely! I knocked it together as a spur of the moment POC after your review on the other PR, so I expected there to be some code quality issues. 

Look forward to your suggestions when you have time 😄  have a great day!

adrienpoly:
General comments 

## option

I would like to keep a consistent pattern for observer mixins with the first parameter being the controller and the second the options. 

So it should be useTargetChanges(this, options)

For passing the targets see my following point

## Targets

To access the controller array of targets you have to pass through the `constructor`. (This also took me some time to figure out). Some examples can be found in `useMemo`, `useDebounce` (https://github.com/stimulus-use/stimulus-use/blob/master/src/use-memo/use-memo.ts#L9)

So here something like `controller.constructor.targets` should return the array of string 🎉 

My suggestion for the options is now that you can access easily the list of targets by default the mixin can observe all of them and as an option pass `{ targets: ["aTargetToWatch"] }` 

| Option| Description |Default value|
|-----------------------|-------------|---------------------|
| `targets` | an array of target names to track for mutations | `all` |

## mixin name

`useTargetChanges` is explicit the `s` makes it a bit different from other mixin name ?

an alternative could be `useTargetMutation`

src/use-target-changes/use-target-changes.ts
```diff
@@ -0,0 +1,158 @@
...
+  deprecatedTargetSelector: string
+
+  constructor(controller: TargetChangesController, targets: string[], options: TargetChangesOptions = {}) {
+    super(controller, options)
+
+    this.targetElement = controller.element
+    this.targets = targets
+    this.deprecatedTargets = this.targets.map(target => `${controller.identifier}.${target}`)
+    this.controller = controller
+    this.deprecatedTargetSelector = `data-target`
```
adrienpoly, line 26: Excellent to track both old and new format.
it would be nice to get it from the `schema` as this is configurable (probably not used much)
reference https://github.com/hotwired/stimulus/issues/124#issuecomment-372449543

src/use-target-changes/use-target-changes.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.targetSelector = `data-${this.controller.identifier}-target`
```
adrienpoly, line 27: for Stimulus 2 this should be ok 
https://github.com/hotwired/stimulus/blob/main/packages/%40stimulus/core/src/target_set.ts#L54
but I think it could change in the future, I remember seeing an issue on this.

Sub-Xaero:
Thanks again for the review. The magic incantation `controller.constructor.targets` was just what I was looking for ✨ 

I've made those suggested changes and cleaned up the code a little bit.

I've tried to clean up the fragmented target selector logic, not sure if you can see any further improvements? 
Rather than assuming an alternate selector of `data-target`, I've pulled it from the schema as you suggested. I've also left a note in about the identifier scoped target selector in the hopes that something like https://github.com/hotwired/stimulus/pull/362 gets merged in, but I suspect there's not much to be done for the time being.

Let me know if I've missed anything 😊

docs/use-target-mutation.md
```diff
@@ -0,0 +1,93 @@
...
+import { Controller } from 'stimulus'
+import { useTargetMutation } from 'stimulus-use'
+
+export default class extends Controller {
+
+  static targets = ["location"]
+
+  connect() {
+    useTargetMutation(this, ["location"])
+  }
```
adrienpoly, line 62: add an example with the `targets` option

```suggestion
  static targets = ["location", "content", "view"]

  connect() {
    useTargetMutation(this, { targets: ["location"] }) // only track mutations of location targets
  }
```

docs/use-target-mutation.md
```diff
@@ -0,0 +1,93 @@
...
+It adds two new behaviors to your Stimulus controller for each target you specify: `[target]TargetAdded` and `[target]TargetRemoved`, which will get triggered when targets get added to your
+controller, either by adding the `data-[controller]-target="[target]"`/`data-target="[controller].[target]"` to an existing element in the controller's scope, or by adding a new element with those
+attributes to the controller's scope.
+
+`[target]TargetAdded` and `[target]TargetRemoved` only receive one parameter, the target node that was added or removed.
+
+## Reference
+
+```javascript
+useTargetMutation(controller, targets, options)
```
adrienpoly, line 14:
```suggestion
useTargetMutation(controller, options)
```

docs/use-target-mutation.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+**controller** : a Stimulus Controller (usually `'this'`)
+
+**targets** : an array of the targets you wish to watch additions and removals of
```
adrienpoly, line 19:
```suggestion
```

src/use-target-mutation/target-mutation-controller.ts
```diff
@@ -0,0 +1,20 @@
...
+    requestAnimationFrame(() => {
+      const [observe, unobserve] = useTargetMutation(this, this.options)
+      Object.assign(this, { observe, unobserve })
+    })
+  }
+
+  observe!: () => {}
+  unobserve!: () => {}
+
+  [index: string]: any;
```
adrienpoly, line 19: Cool this is to somehow declare dynamic callbacks in TS?

adrienpoly (commented):
A few minor corrections to the doc and suggestions

Do you think it would be easy to add the **name**TargetChanged calback too?

docs/use-target-mutation.md
Sub-Xaero: Good spot!

src/use-target-mutation/target-mutation-controller.ts
Sub-Xaero: Kinda, that's what enabled this line:
```js
this.controller[targetCallback] && method(this.controller, targetCallback).call(this.controller, node)
```
Tells Typescript that the class accepts indexing its properties via string, and what type of thing is returned when you do that. 

https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures

If you decide to continue with https://github.com/stimulus-use/stimulus-use/pull/105 and just using `method(this.controller, targetCallback)`, then the need to have this declared goes away.

docs/use-target-mutation.md
Sub-Xaero: Good spot!

Sub-Xaero:
I've made those changes, thanks for that!

What would you expect the behaviour of `nameTargetChanged ` to be?

adrienpoly:
> What would you expect the behaviour of nameTargetChanged to be?

I was thinking of a target being updated by a turbo stream but as Turbo performs a full replacement I guess this would trigger a remove/add callback cycle anyway.

Maybe it could make more sense for StimulusReflex where there is a real morph update being performed?

Anyway It is looking perfect to me let's ship it and see if they are real behavior coming up to support such feature

Sub-Xaero:
@adrienpoly Ah that makes sense, yes that would be useful. Have a look, I just added this feature on my lunch break, managed to clean up a lot of the duplication because of it. Tried to keep it as lean as possible so that big DOM operations don't run slowly 🤞  - but haven't been able to test extensively.

Without knowing too much about how cable_ready and StimulusReflex morphs translate into DOM operations, I can't say for certain that it will always be sensibly reported - entirely relies on the browser reporting changes correctly. Tested it in a basic `cable_ready` sandbox, seems to work well with my toy example. 

Let me know what you think.

leastbad:
On the bright side, it sure seems like we prompted them into action.

Sub-Xaero:
Did I miss something? What's occurring?

marcoroth:
@Sub-Xaero This PR here: https://github.com/hotwired/stimulus/pull/367

Sub-Xaero:
@marcoroth Oooh slick, glad that's becoming a core feature. Let's hope it gets merged in soon!

adrienpoly:
Will merge this one anyway just need to play a bit more with it. the api seems to be the same so it will be plug and play with 2022 2.1 release 😀

adrienpoly:
Thanks @Sub-Xaero 🚀

Sub-Xaero:
@adrienpoly My pleasure! Hope it's useful to people 😁

PR #105, opened 2021-01-18
Fix typing information for pre-declared methods that are added by the enhancement functions

Sub-Xaero:
Hiya, 

Using a few of these in my Typescript projects. By describing the methods like they are currently
```
options!: *Functionality*Options
observe!: () => void
unobserve!: () => void
```
extending the controllers works fine, but it throws errors if a controller tries mixing in `use*Functionality*` without implementing all of them. 

Might just be my usage of them, but I found myself having to declare `observe` and `unobserve` on my controller even though I don't use them. And unless the user is extending the class - having a class property called `options` should be optional, given that mixin method allows passing inline options.

I've made changes so that the "extending" mode properties are not expected on the mixin function call signatures. 

---

Also fixed the type of `setTimeout` in the `use-debounce`, 

---

Let me know what you think.

### DIFF
--- a/src/support/index.ts
+++ b/src/support/index.ts
@@ -4,8 +4,9 @@ export const method = (controller: Controller, methodName: string): Function => 
   const method = (controller as any)[methodName]
   if (typeof method == 'function') {
     return method
+  } else {
+    return (...args: any[]) => {}
   }
-  throw new Error(`undefined method "${methodName}"`)
 }
 
 export const composeEventName = (name: string, controller: Controller, eventPrefix: boolean | string) => {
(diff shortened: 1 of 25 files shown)


adrienpoly:
Thanks this was very much needed indeed !

Do you see an alternative way to avoid this `Omit` syntax ? I would like to keep the signature of the useFunction as clean as possible.

```ts
export const useHover = (controller: Omit<HoverController, 'options' | 'observe' | 'unobserve'>, options: HoverOptions = {}) 
```

I will look into it deeper, I made a few quick tests, and it looks like a few things could be simplified/corrected at the same time.

### method function 

the `method` function is call with a safeguard  

```js
this.controller.invisible && method(this.controller, 'invisible').call(this.controller, event)
```
This safeguard requires the controller to declare the type for the `invisible` method but inside `method` we test if the method exist or not (job is done twice). I think this code somehow derived from its initial intent, and it is good to look back into it.

a simplified version could be to return a noop function if not found this removes all safe gard and associated type

```js
export const method = (controller: Controller, methodName: string): Function => {
  const method = (controller as any)[methodName]
  if (typeof method == 'function') {
    return method
  } else {
    return noop
  }
}
```

### Controller variable

The other requirement for type is the variables within the controller.

for useVisibility for example, it is `isVisible`

a solution could be to define a `UseVisibilityInterface` 

```js
import { Controller } from "stimulus"

export interface UseVisibilityController extends Controller {
  isVisible: boolean
}
```

and then 

```js
constructor(controller: UseVisibilityController, options: VisibilityOptions = {}){
  ....
}
```

WDYT? From preliminary tests it looks like this could largely simplify the useFunction signature and avoid `Omit`

Sub-Xaero:
Absolutely yeah. We could split the delaration, into something like 
```js
import { Controller, Context } from 'stimulus'
import { useResize, ResizeOptions } from './use-resize'

export class ResizeComposableController extends Controller {
  declare resize: (contentRect: DOMRectReadOnly) => void
}

export class ResizeController extends ResizeComposableController {
  options?: ResizeOptions

  constructor(context: Context) {
    super(context)
    requestAnimationFrame(() => {
      const [observe, unobserve] = useResize(this, this.options)
      Object.assign(this, { observe, unobserve })
    })
  }

  declare observe: () => void
  declare unobserve: () => void
}

export const useResize = (controller: ResizeComposableController, options: ResizeOptions = {}) => {
...
}
```


So the exported type that users can extend from has all the properties, including `observe`, `unobserve`, but the `useFunction` is kept nice and clean and just requires the callbacks and variables to be defined.

PR #106, opened 2021-01-18
Fix typo in ThrottleController, "trottles" instead of "throttles"

Sub-Xaero:
Hiya, 

Small typo fix I found when I was working on https://github.com/stimulus-use/stimulus-use/pull/105


adrienpoly:
good catch 🙏

PR #111, opened 2021-02-26
use-transition: remove active-to class when starting leave, and vice-versa

### LINKED ISSUE #110: useTransition: "enterTo" remains during "leave" (and vice versa on "enter")
Hi there!

Thanks for the wonderful library - i'm building a tutorial with that uses it right now - it's awesome!

I'm testing out the beta transitions. I'm not using Tailwind in this app, so the enter-leave classes are not the same as the sandbox. I've noticed a slight issue. For example, suppose we have:

```js
    connect() {
        useClickOutside(this);
        useDebounce(this);
        useTransition(this, {
            element: this.resultTarget,
            enter: 'transition-fade-enter',
            enterActive: 'transition-fade-enter-active',
            enterTo: 'transition-fade-enter-to',
            leave: 'transition-fade-leave',
            leaveActive: 'transition-fade-leave-active',
            leaveTo: 'transition-fade-leave-active-to',
        });
    }
```

The `enter()` works great, the element looks like this after:

```html
<div
    class="transition-fade-enter-to"
    data-search-preview-target="result"
> ....</div>
```

However, when I `leave()`, the `enterTo` class (which is `transition-fade-enter-to`) *remains* on the element permanently. For example, once `leave()` finishes, the HTML looks like this:

```html
<div
    class="transition-fade-enter-to transition-fade-leave-active-to hidden"
    data-search-preview-target="result"
> ....</div>
```

Basically, both the `enterTo` and `leaveTo` classes are "stuck" on there permanently. This kills all future transitions because, for example,  if `transition-fade-enter-to` has `opacity: 1` that overrides the `opacity 0` set by `transition-fade-leave-active-to`. 

I think the solution is "simply" to, on leave, remove the `enterTo` class if it's there. And on enter, remove the `leaveTo` class if it is there.

Let me know if that makes sense :).

Cheers!

weaverryan:
Hi!

This fixes #110 - I did try this locally.

Please let me know if you see any issues or potential pitfalls. 

Thanks!

### DIFF
--- a/src/use-transition/use-transition.ts
+++ b/src/use-transition/use-transition.ts
@@ -59,12 +59,13 @@ export const useTransition = (controller: TransitionComposableController, option
     const enterClass = getAttribute("enter", options, dataset)
     const enterActiveClass = getAttribute("enterActive", options, dataset)
     const enterToClass = getAttribute("enterTo", options, dataset)
+    const leaveToClass = getAttribute("leaveTo", options, dataset)
 
     if (!!hiddenClass) {
       targetElement.classList.remove(hiddenClass)
     }
 
-    await transition(targetElement, enterClass, enterActiveClass, enterToClass)
+    await transition(targetElement, enterClass, enterActiveClass, enterToClass, leaveToClass)
 
   }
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Or should the `*to` class be removed once the transition is finished (I know Vue does this)? So once the "enter" is finished, all of the transition classes would be removed (including `activeTo`).

adrienpoly:
@weaverryan I think I have fixed your issue in #115. Sorry I didn't used your PR as a starting pointed but I needed to do quite a few changes internally to addressed it. 

I have release the 0.24 beta version if you could try it and let me know (yarn add stimulus-use@beta ). 

I will still work a bit on the documentation and also make all options available as directives

PR #114, opened 2021-03-02
Use transition test suite foundation

adrienpoly:
very basic useTransition test suite foundation before addressing #110 and #111


weaverryan:
Awesome idea 👍

adrienpoly:
@weaverryan i made significant progress, I should have a version to test by the end of the day

PR #124, opened 2021-03-20
log correct event name in `dispatch()` when using `eventPrefix`

marcoroth:
I noticed that when using a custom `eventPrefix` that the `dispatch()` function logs the wrong `eventName`. 

If you check the log messages in the dev tools and expand the details you will see the the the `eventName` is logged without the prefix while the right eventName (with the prefix) is actually emitted. 

This PR fixes the log output to match the actual dispatched event.


adrienpoly:
Thanks @marcoroth 

@all-contributors please add @marcoroth to bug

PR #128, opened 2021-03-24
Introduce `useMatchMedia` mixin

marcoroth:
This PR introduces the `useMatchMedia` mixin as discussed in #125.

After writing this mixin I'm not sure if we should even add the `useBreakpoints` mixin. But on other hand I can still see them co-exist for more simpler and just width-based use-cases. But I feel like this one is very exiting!

Thoughts, comments and opinions are very welcome!

## Example

```js
import { useMatchMedia } from 'stimulus-use'

  connect() {
    useMatchMedia(this, { 
      mediaQueries: {
        'small': '(min-width: 320px) and (max-width: 769px)',
        'tall': '(min-height: 1000px)',
        'light': '(prefers-color-scheme: light)',
        'landscape': '(orientation: landscape)',
      }
    })
  }

  smallChanged(event) { /* ... */ }
  isSmall(event)      { /* ... */ }
  notSmall(event)     { /* ... */ }

  // The same for 
  // tall      => tallChanged()      // isTall()      // notTall
  // light     => lightChanged()     // isLight()     // notLight()
  // landscape => landscapeChanged() // isLandscape() // notLandscape()          
}
```

### Still To-Do

* [x] Initial callback on `connect()`
* [x] Docs
* [x] More options
* [x] Tests


marcoroth:
I was wondering if it would be better to define the media queries as static values, like as Stimulus does with the targets/values and as we do with a lot of other mixins. Something like:

```js

  static mediaQueries = {
    small: '(min-width: 320px) and (max-width: 769px)',
    tall: '(min-height: 1000px)',
    light: '(prefers-color-scheme: light)',
    landscape: '(orientation: landscape)'
  }

  connect() {
    useMatchMedia(this)
  }
}
  ````

**Comment by adrienpoly** [comment]
@marcoroth lets keep the media queries as an option of `useMatchMedia` for now. there is still this on going [discussion](https://github.com/stimulus-use/stimulus-use/discussions/64) where we were leaning toward direct options rather than using a static object.

**Comment by marcoroth** [comment]
Fine for me. I don't really have a preference, just wanted to bring it up 👍

**Comment by adrienpoly** [comment]
@pascallaliberte Thanks for this proposal. It is interesting, I would like to avoid having a dependency with Tailwind but If I understand correctly your proposal is more generic and the idea is to have a helper that would take `breakpoints` and convert them to `mediaQueries`

Example

```js
const breakpoints = {
  "sm": "640px",
  "md": "768px",
  "lg": "1024px",
  "xl": "1280px",
  "2xl": "1536px"
}

function breakpointsToMediaQueries(breakpoints) {
...
}

console.log(breakpointsToMediaQueries(breakpoints))
{
  sm: '(min-width: 0px) and (max-width: 640px)',
  md: '(min-width: 641px) and (max-width: 768px)',
  ...
}
```

Am I understanding correctly?

pascallaliberte:
@adrienpoly gotcha about Tailwind.

Yep, that approach would do the trick.

Now mind you, to be more _extractive_ rather than _predictive_ about this feature, here's more context: I came close to building something similar on my end. I considered importing the Tailwind configs, and in the end I hesitated because I didn't want to risk the extra work of potentially [having to reduce the payload, as Tailwind's documentation suggests needing to do](https://tailwindcss.com/docs/configuration#referencing-in-java-script). In the end, I created a smaller `page-layout` controller that just creates `(min-width: ${breakpoint}px)` media queries and simply dispatches a `page-layout:changed` when breakpoint thresholds are crossed. The controllers that are subscribing to that event simply evaluate the `window.innerWidth` when catching the event, and determine what to do about it. This allows more fine-grained breakpoints, easily passed as values e.g. `data-page-layout-breakpoints-value="<%= [638, 768].to_s %>".

No doubt Tailwind's breakpoints will be the default for many folks (your suggested helper will be useful), but that was my story of shying away from them.

evenreven:
Just a (somewhat worthless) comment to say that I can't wait to start using this once it's merged. I have some javascript that's not supposed to run on smaller screens, so it would really make my controller so much better tidier to have something like this. 👍

(Sadly, I'm useless at writing tests and the like, so I can't help it along.)

adrienpoly:
@marcoroth is this still a draft ? It looks good to me

marcoroth:
@adrienpoly It's still draft because it needs docs and tests.

Be if we can agree on the API as shown in the PR description I can finish the remaining todos.

adrienpoly:
I have a final thought on the naming what about replacing `Changed` by `Media`

```js
  smallMedia(event) { /* ... */ }
  isSmall(event)      { /* ... */ }
  notSmall(event)     { /* ... */ }

  // The same for 
  // tall      => tallMedia()      // isTall()      // notTall
  // light     => lightMedia()     // isLight()     // notLight()
  // landscape => landscapeMedia() // isLandscape() // notLandscape() 
```

WDYT?

marcoroth:
@adrienpoly I'm wondering if this isn't causing more confusion. 

The `Changed` callback is called when the media query gets active and when it gets inactive. I'm not sure if you get that with the name `Media`.

evenreven:
If it's a callback for _changes_ in both directions, I (as a future user of this method) would have to side with @marcoroth on this one. `Media` doesn't signify that it's a change, imho. `Changed`, on the other hand, does. 

That said, maybe I'm just lacking in imagination, but what's a use case for a combined enter/leave callback in this case? Toast notifications? `classList.toggle`s? Isn't it more robust anyway to use the callbacks based on the actual state, not whether or not it changed in one direction or the other? Right now I can only see the million things that could go wrong using this (like applying changes based on the wrong mq), but maybe I'm just not getting it completely! :)

This is just my two cents - which, since I'm not developing this, you can (and probably should!) disregard completely!

marcoroth:
I think this is ready. Feel free to test if this works as expected and to give feedback if not! 😊 

@adrienpoly Looks like the Circle CI setup is still missing the ENV variables

meneerprins:
Are there still plans to merge this PR? This feature would be really useful!

adrienpoly:
sorry for not merging this earlier was kind of busy recently. this is awesome work @marcoroth. And thanks all for your inputs

PR #130, opened 2021-04-13
Enabling SauceLab IE11 config

Matts:
Testing for #3

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -42,7 +42,7 @@ const customLaunchers = {
     browserName: 'safari',
     platform: 'macOS 10.13',
     version: 'latest'
-  }
+  },
   // sl_edge_17: {
   //   base: 'SauceLabs',
   //   browserName: 'microsoftedge',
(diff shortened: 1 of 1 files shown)


Matts:
I think you need to trigger the pipeline manually seeing as it is coming from an external fork? If issue is reproducable I will create a new ticket

adrienpoly:
Ok this is very strange the CI used to show up on PRs... I tried to change a setting. Anyway I ran manually the CI and yes it seems to be reproducible as all test on IE11 fails. 
However, I haven't really tuned the bundler for the test environment neither at this time.

Matts:
Personally I find it weirder that my own babel configuration has no issues with classes, seeing as Stimulus controllers work fine. Just for some reason it fails to mangle the stimulus-use source. Do you maybe have a setting that denies babel?

Maybe I need to install a ts plugin?

adrienpoly:
If you open the npm package you can see the dist versions. I know that the dist version is fairly light in terms of transpilation so there is still probably the `Class` and `const` in it.  What is strange is that as you said it doesn't seems to be transpiled correctly.

Matts:
It seems that this is intended logic for my app configuration. Babel does not compile javascript in node_modules. I am using Webpack Encore and so apparently (thanks @weaverryan) can use `includeNodeModules` 
https://github.com/symfony/webpack-encore/blob/8c2752afd4d63d0e72c757b96af6ae2c2d3c85cf/index.js#L891

About your project not compiling to es5. Maybe because your target is es2019?

adrienpoly:
> About your project not compiling to es5. Maybe because your target is es2019?
I remember that I had to do this to keep the library full tree shakable, but I don't remember all the details and was assuming that bundlers always make the final compilation.

I am not really familiar with Symfony Encore I suppose it is the equivalent to Webpacker in Rails. Your results were with a production build or dev? In Webpacker in development node_module packages are not compiled, but they are for production (and treeshake). 

for your suggestion of documenting this yes this is a great idea. How about creating a new section for Polyfills. This section can list the polyfills listed in #3 as well as other tips for older browser

adrienpoly:
closing this for now. Will reopen a dedicated PR to extend the CI and test build to other environements

PR #132, opened 2021-04-21
Don't try to process getAttribute on comment nodes

Intrepidd:
I realised I got errors when using target mutation because some library on my page adds HTML comments, this should fix it


adrienpoly:
@all-contributors please add @Intrepidd  to bug

PR #144, opened 2021-05-17
Add `isConnected` helper to `ApplicationController`

marcoroth:
### `isConnected` helper

This PR adds a `isConnected` helper function the the `ApplicationController`.

This helper function can be used to detect whether a Stimulus controller is connected or not.

#### Example

```js
import { ApplicationController } from 'stimulus-use'

export default class extends ApplicationController {
  initialize() {
    console.log("initialize", this.isConnected) // => false
  }

  connect() {
    console.log("connect", this.isConnected) // => true
  }

  anyOtherAction() {
    console.log("anyOtherAction", this.isConnected) // => true
  }

  disconnect() {
    console.log("disconnect", this.isConnected) // => false
  }
}
```

### DIFF
--- a/docs/application-controller.md
+++ b/docs/application-controller.md
@@ -27,6 +27,7 @@ export default class extends ApplicationController {
 
   connect() {
     this.isPreview // true/false if it is a Turbolinks preview
+    this.isConnected // true/false if the controller is connected
     this.dispatch("hello") // helper to dispatch a custom event "greet:hello" to other Stimulus controllers
   }
 }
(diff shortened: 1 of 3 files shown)


leastbad:
I am really excited to see this land. It solves a blind-spot in Stimulus when using the Values API.

When you have a value set via data attributes in your markup, Stimulus fires a `ValueChanged` callback after `initialize` but before `connect`.

This behaviour is highly desirable about 60% of the time; you can use it to initialize values before the controller connects. However, sometimes you do not want the `ValueChanged` callback to fire before it is connected, and you have to manually `return`.

Roughly 15% of all my controllers have code very similar to the following:

```js
initialize() {
  this.connected = false
}
connect() {
  this.connected = true
}
disconnect() {
  this.connected = false
}
completedValueChanged() {
  if (!this.connected) return
}
```

This PR would allow me to remove all of the ceremonial code (and the instance variable!) and just do this:

```js
completedValueChanged() {
  if (!this.isConnected) return
}
```

leastbad (approved):
Now, it's finally ready 😉

adrienpoly:
Thanks @marcoroth I wasn't aware of this usecase glad to have this new helper

PR #148, opened 2021-06-08
Add use-hotkeys behavior

julianrubisch:
This PR adds a behavior to proxy the (un)registering of Hotkeys to the hotkeys-js library, which has been added as a peer dependency.

Tests and documentation have been updated accordingly.

Example of a config, taken from the tests:

```js
connect() {
    useHotkeys(this, {
      hotkeys: {
        '/': {
          handler: 'singleKeyHandler'
        },
        'cmd+a': {
          handler: 'metaKeyHandler'
        },
        f: {
          handler: 'scopeHandler',
          options: {
            scope: 'files'
          }
        },
        b: {
          handler: 'inputHandler'
        },
        c: {
          handler: 'keyUpHandler',
          options: {
            keydown: false,
            keyup: true
          }
        },
        'ctrl-d': {
          handler: 'splitKeyHandler',
          options: {
            splitKey: '-'
          }
        }
      },
      filter: this.filter
    })
}
```


src/use-hotkeys/use-hotkeys.ts
```diff
@@ -0,0 +1,17 @@
+import { Controller } from 'stimulus'
+// @ts-ignore
+import hotkeys from 'hotkeys-js'
+
+export const bindHotkeys = (controller: Controller) => {
```
marcoroth, line 5: I think we should keep the `use` naming scheme throughout the library. Probably `useHotkeys` would fit better?

spec/use-hotkeys/use_log_controller.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default class UseLogController extends Controller {
+  static hotkeys = { '/': 'singleKeyHandler', 'cmd+a': 'metaKeyHandler' }
+
+  connect() {
+    bindHotkeys(this)
+  }
+
+  disconnect() {
+    unbindHotkeys(this)
```
marcoroth, line 12: Does the developer really need to call this manually? Can't we somehow attach a callback/eventlistener on document or somewhere to call this function accordingly?

src/use-hotkeys/use-hotkeys.ts
julianrubisch, line 5: We might want to provide a second option to define whether or not to [filter](https://github.com/jaywcjlove/hotkeys#filter)

spec/use-hotkeys/use_log_controller.js
```diff
@@ -0,0 +1,22 @@
+import { Controller } from 'stimulus'
+import { bindHotkeys, unbindHotkeys } from '../../src'
+
+export default class UseLogController extends Controller {
+  static hotkeys = { '/': 'singleKeyHandler', 'cmd+a': 'metaKeyHandler' }
```
julianrubisch, line 5: It could make sense to think about a way to declare [options](https://github.com/jaywcjlove/hotkeys#option)

docs/use-hotkeys.md
```diff
@@ -0,0 +1,47 @@
...
+## Usage
+
+**Composing**
+
+```js
+import { Controller } from 'stimulus'
+import { bindHotkeys, unbindHotkeys } from 'stimulus-use'
+
+export default class extends Controller {
+  static hotkeys = {'cmd+a': 'handler'}
```
marcoroth, line 33: We somewhat agreed that we shouldn't introduce more `static` configurations.

It would also makes sense here. Instead you would pass the options directly with the `bindHotkeys` function.

Some like: 

```js
bindHotkeys(this, {
  hotkeys: {
    'cmd+a': 'handler'
  }
}
```

This would also allow for other options to be specified.

marcoroth (commented):
Hey @julianrubisch!

Thank you for this PR, this is looking awesome so far!

I've added a few comments with suggestions in order to make all behaviors feel more or less similar.

Let me bow what you think!

spec/use-hotkeys/use_log_controller.js
julianrubisch: If you know of any way to do this, that would be 👨‍🍳 😘 ... I currently do not. That would also solve the naming problem above.

@adrienpoly ?

docs/use-hotkeys.md
julianrubisch: Ah, didn't know that. If that is the general contract, I'll happily refactor.

spec/use-hotkeys/use_log_controller.js
julianrubisch: Do stimulus controller emit an event on disconnect that I could use here? But then again, that's another event listener 🤔

julianrubisch: Ah, I'll take a look at use-hover for example, that looks sane 👍🏻

julianrubisch:
Redone! Somehow the tests are failing on CircleCI, I assure you they're all green locally 🤷‍♂️

leastbad:
Forgive me - reading TypeScript is not yet native to me, so I need to ask for clarification:

Are you specifying the methods to receive the key event with strings? Does `handler` also accept functions? eg.
```js
handler: 'foo'
// or
handler: this.foo
```
Are those equivalent?

julianrubisch:
good question because I've just taken this from how it worked before... might be worth a try!

julianrubisch:
> Forgive me - reading TypeScript is not yet native to me, so I need to ask for clarification:
> 
> Are you specifying the methods to receive the key event with strings? Does `handler` also accept functions? eg.
> 
> ```js
> handler: 'foo'
> // or
> handler: this.foo
> ```
> 
> Are those equivalent?

So I've fiddled around with it a bit, and it requires binding to `this`, and for the handlers to actually conform to the `KeyHandler` type.

I didn't _quite_ get it working, but it actually looked uglier and I think would put too much burden on the developer...

leastbad:
Do they have to conform even though most Stimulus controllers are not written in TypeScript?

Usually, I get around `this.foo.bind(this)` by using a trick @joshleblanc taught me:
```js
foo = event => {}
```
For reasons I could not explain while drinking, `this` works when using the assignment form instead of the function form.

julianrubisch:
Agreed. I will try to keep pushing, however at the moment, at least we've got a working solution

julianrubisch:
Done!

adrienpoly:
> Redone! Somehow the tests are failing on CircleCI, I assure you they're all green locally 🤷‍♂️
> 
> Edit:
> 
> Ah! `Make sure the SAUCE_USERNAME and SAUCE_ACCESS_KEY environment variables are set.` ?!

@julianrubisch I need to look into this. I check several times and my secrets are correct, not sure what is missing

Just starting to read your PR it looks awesome

marcoroth:
I just chatted with @julianrubisch and we had some other ideas on how we could approach this mixin.

Instead of passing all hotkeys to one `useHotkeys` call we could do one-offs:

```js
useHotkeys(this, '/', this.singleKeyHandler.bind(this))
useHotkeys(this, 'cmd+a', this.metaKeyHandler.bind(this))
```

I'm sure we can also remove the `.bind(this)` since we explicitly pass `this` to `useHotkeys`. So we could probably handle that in the library itself.

----

Another idea would be to use "names" for the hotkeys:

```js
connect() {
  useHotkeys(this, '/', 'overlay')
}

overlayPressed(event)   { /***/ }
overlayReleased(event)  { /***/ }
overlayChanged(event)   { /***/ }
```

This would implicitly call the callback functions based on the given name. This would be close to what you know from Stimulus (see Values API).

----

If you want to define multiple hotkeys in one call it could look like this:

```js
  useHotkeys(this, {
    hotkeys: {
      '/': 'overlay',
      'cmd+a': 'keyhandler'
    }
  })
```

---

And if you want to pass in options for a more advanced use case it could look like this:

```js
connect() {
  useHotkeys(this, {
    hotkeys: {
      '/': {
        handler: this.overlayPressed
        // other options
      },
      'cmd+a': {
        handler: this.keyhandlerPressed
        // other options
      }
    }
  })
}
```

---

With that we cover a wide range of use cases and it's highly customizable. But if you just want a single hotkey it's as easy as dropping in a one liner:

```js
useHotkeys(this, '/', 'overlay')
```

Let me know what you think!

julianrubisch:
✅ implicit `this` binding
✅ "simple mode"

julianrubisch:
Anything I can do to advance this?

julianrubisch:
Hold on, I found an issue with the simple syntax. In fact calling `useHotkeys` multiple times won't work because it's not idempotent, overwriting `connect` and `disconnect` multiple times (on top of the global stuff that `hotkeys` does)

Here's a suggestion for a compromise:

```js
useHotkeys(this, { '/': [this.eventHandler], 'a': [this.inputHandler, this.inputTarget]})
```

julianrubisch:
Implemented that compromise and updated the docs, please let me know what you think 🙏

adrienpoly:
this is awesome @julianrubisch and thanks @marcoroth, @leastbad for the reviews

PR #151, opened 2021-06-11
Enforce Prettier formatting

marcoroth:
I noticed that we have a `.prettierrc.json` and that it didn't run in a while.

With that I added `prettier` as a `devDependency` and added a GitHub Action to check the formatting automatically.

If you think that doesn't make sense feel free to close 😊


adrienpoly:
make a lot of sense I will just merge all other PR first and run prettier on the main branch to reduce the merge conflicts

PR #153, opened 2021-06-12
Introduce `useWindowFocus` mixin

marcoroth:
This PR introduces the `useWindowFocus` mixin which tracks if the window is focused or not.

While `useVisibility`  just checks if the window might be visible `useWindowFocus` actually checks if the window is focused or not.

An example might look like this:
```js
// video_controller.js

import { useWindowFocus } from 'stimulus-use'

  connect() {
    useWindowFocus(this)

    this.player = new VideoPlayer()
  }

  focus() {
    this.player.play()
  }

  unfocus() {
    this.player.pause()
  }
}
```

I'm open to feedback. Make sure to let me know if you disagree with something! 😊


leastbad:
In layman's terms, you're saying that this knows if it's the active tab AND if the browser has focus? vs. just being the active tab but behind other windows...?

marcoroth:
@leastbad That's exactly right!


https://user-images.githubusercontent.com/6411752/121771957-c8c50f00-cb72-11eb-8714-73ed57aaa067.mov

leastbad:
There's everything to love about this! My only question is whether it makes sense to "fix" `useVisibility` or keep them separate. Not advocating either way, but I do wonder who wants one but not the other. The "why" of these functions is "is the user engaged right now?"

marcoroth:
That's a reasonable question to ask. I first thought about expanding the `useVisibility` mixin but then decided to extract it into a separate one. The two mixins use different browser APIs ([`Document.visibilityState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) vs. [`Document.hasFocus`](https://developer.mozilla.org/en-US/docs/Web/API/Document/hasFocus)) and serve for slightly different use cases. 

That's why I would keep them as two separate mixins. Maybe we could make it more clear by using different and more meaningful names? But I have no hard feelings for one or the other.

leastbad:
I like the names and I think your reasoning holds water. 👍 from me! I cannot wait to put this to great use.

PR #159, opened 2021-06-24
chore: change "main" to a module instead of UMD.

ParamagicDev:
@adrienpoly I havent had the chance to test this as im currently quite limited on time, but give this a whirl and let me know if this fixes the issues you were facing with tree-shaking.


adrienpoly:
thanks @ParamagicDev will look into this direction

adrienpoly:
@all-contributors please add @ParamagicDev to code

PR #163, opened 2021-07-30
clean: remove redundant targetElement property

Vafilor:
The StimulusUse base class already provides this property.


adrienpoly:
thanks

Issue #209, opened 2022-07-11
Introduce useRequestIdleCallback optimization hook


gremo:
Why not introduce `useRequestIdleCallback` to use [`window.requestIdleCallback()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) easily in controllers?

> Queues a function to be called during a browser's idle periods. This enables developers to perform background and low priority work on the main event loop.

[Can I use `requestIdleCallback`](https://caniuse.com/requestidlecallback)
 
<details>
  <summary>WIP hook example in JavaScript</summary>
  
```js
/**
 * @typedef {import('@hotwired/stimulus').Controller} Controller
 *
 * @typedef {Object} HookOptions
 * @property {number} [maxWait]
 */

const defaultOptions = {
  maxWait: undefined,
};

/**
 * @param {Controller} controller
 * @param {HookOptions} options
 * @returns {[function(): void, function(): void]}
 */
export const useRequestIdleCallback = (controller, options) => {
  const {
    maxWait,
  } = { ...defaultOptions, ...options };

  const idleCallbackIds = [];

  const idling = (fn, timeout) => {
    if (requestIdleCallback) {
      return function (...args) {
        const idleCallbackId = requestIdleCallback(deadline => fn.call(this, ...args, deadline), { timeout });

        idleCallbackIds.push(idleCallbackId);
      }
    }

    const start = Date.now();

    return function (...args) {
      const timeoutId = setTimeout(() => {
        fn.call(this, ...args, { didTimeout: false, timeRemaining: () => Math.max(0, 50 - (Date.now() - start)) });
      }, 1);

      idleCallbackIds.push(timeoutId);
    };
  };

  const cancelIdling = id =>
    (cancelIdleCallback || clearTimeout)(id);

  const observe = () => {
    controller.constructor.idles?.forEach(func => {
      if ('string' === typeof func) {
        controller[func] = idling(controller[func].bind(controller), maxWait);
      }

      if ('object' === typeof func) {
        const { name, maxWait } = { ...func, ...options };
        if (!name) {
          return;
        }

        controller[name] = idling(controller[name].bind(controller), maxWait);
      }
    });
  };

  const unobserve = () => {
    idleCallbackIds.forEach(id => cancelIdling(id));
  };

  const controllerDisconnect = controller.disconnect.bind(controller);

  Object.assign(controller, {
    disconnect() {
      unobserve();
      controllerDisconnect();
    },
  });

  observe();

[...]

PR #212, opened 2022-07-26
Adds a new hook useBackgroundQueue

gremo:
Allow use of window.requestIdleCallback()` APIs easily in controllers. See #209


docs/use-background-queue.md
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+## Per function `maxWait` option
+
+If you need to set a different wait option for each function, you can specify it within the `queue` array like so:
+
+```js
+export default class extends ApplicationController {
+  static debounces = [
```
marcoroth, line 45: This is to make it match the other one
```suggestion
  static queue = [
```

but I'm also not sure if this name isn't a bit misleading. I'd vote for `static idleCallbacks = []`

src/use-background-queue/use-background-queue.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface BackgroundQueueOptions {
+  maxWait?: number,
+  name?: string
+}
+
+const defaultOptions = {}
+const idleCallbackIds: number[] = []
+
+const ric = (fn: Function, maxWait?: number) => {
```
marcoroth, line 15: I think we should define `window.requestIdleCallback` if it's not defined already instead of wrapping it in a function and returning the function.

As it's outlined in the article you mentioned [here](https://developer.chrome.com/blog/using-requestidlecallback/#checking-for-requestidlecallback).

src/use-background-queue/use-background-queue.ts
```diff
@@ -0,0 +1,81 @@
...
+    const args = arguments
+    const context = this
+
+    const idleCallbackId = window.requestIdleCallback(deadline => fn.call(context, ...args, deadline), { timeout: maxWait })
+
+    idleCallbackIds.push(idleCallbackId)
+  }
+}
+
+const cancelRic = (id: number) =>
```
marcoroth, line 42: The same for the `cancelIdleCallback`. Also, I would check for `window.cancelIdleCallback` instead of `window.requestIdleCallback` here

marcoroth (changes requested):
Hey @gremo, thanks for opening this!

This is already looking good! I think we should stick closer to the actual API name, because otherwise it might get kinda confusing. I would probably go with something like `useRequestIdle()` or `useRequestIdleCallback()`.

I also left a few comments

PR #215, opened 2022-08-05
Migrate docs to `vitepress`

marcoroth:
I just put up this PR to play around with `vitepress`.  The project is still in alpha and has some small issues.

This PR doesn't need to get merged, just thought I put it up here if it ends up being useful in any way.

Demo: https://stimulus-use.netlify.app


henrikbjorn:
That site is tons better than the current one.

marcoroth:
I also like Vitepress a lot. Wondering if @adrienpoly has something against it or if he wants to keep the existing one.

PR #216, opened 2022-08-06
Set `controller` parameter type of all `use` functions to `Controller`

### LINKED ISSUE #213: `useIdle`: Typescript error: argument of type 'this' is not assignable to parameter of type "IdleComposableController"
Because of [this declaration](https://github.com/stimulus-use/stimulus-use/blob/main/src/use-idle/idle-controller.ts#L5) in `IdleComposableController `, there is the following Typescript error:

> Argument of type 'this' is not assignable to parameter of type 'IdleComposableController'.
  Property 'isIdle' is missing in type 'default' but required in type 'IdleComposableController'.

```typescript
import { useIdle } from 'stimulus-use';

export default class extends Controller<HTMLElement> {
  connect() {
    useIdle(this);
  }
}
```

I don't know why that property exists in the first place. It's being setted on:
- https://github.com/stimulus-use/stimulus-use/blob/main/src/use-idle/use-idle.ts#L35
- https://github.com/stimulus-use/stimulus-use/blob/main/src/use-idle/use-idle.ts#L47

Could be solved by `declare isIdle: boolean`... is this my bad or the intended way?

marcoroth:
**Before:**

When using `stimulus-use` with TypeScript it complains if you pass in a regular controller into a `use` function which had the sub-controller as the argument type. 

The type for the first argument was not consistent over all the `use` functions, half of them were `Controller` (which is the "right" way) and the other half was `[Name]ComposableController` (or similar).

Since this is just an issue when using TypeScript in the host-application it probably went unnoticed until now.

```ts
function useIdle(controller: IdleComposableController, ...) {
  // ...
}

  connect() {
    useIdle(this)
    // ^^^^^^^^^^
    // Argument of type 'this' is not assignable to parameter of type 'IdleComposableController'.
    // Property 'isIdle' is missing in type 'default' but required in type 'IdleComposableController'.
  }
}
```

**After:**

This PR changes all `use` functions to take a regular `Controller` as the first argument and casts it to the sub-type in the function body so that the rest of the function can operate on the casted controller.

```ts
function useIdle(composableController: Controller, ...) {
  const controller = composableController as IdleComposableController
  // ...
}

  connect() {
    useIdle(this)
  }
}
````

I'm not super happy with the name `composableController` as the first argument name. If you have any better ideas let me know.

Resolves #213


gjtorikian:
@marcoroth Any chance for a merge and release? Thanks!

marcoroth:
> @marcoroth Any chance for a merge and release? Thanks!

Hey @gjtorikian, yes! I want to cut a new release this week, including this PR.

gjtorikian:
Thank you!

PR #217, opened 2022-08-06
Fix `useIntersection` for multiple elements in the same controller

### LINKED ISSUE #211: `useIntersection` for multiple elements not calls `disappear`
Controller example:
```js
import { useIntersection } from 'stimulus-use';

class ListController extends Controller {
  static targets = ['item'];

  connect() {
    for (const item of this.itemTargets) {
      useIntersection(this, { element: item, root: this.element });
    }
  }

  appear() {
    console.log('✅ appear');
  }

  disappear() {
    console.log('❌ disappear');
  }
}
```

Example: https://stackblitz.com/edit/js-1pzvcb?file=index.js

https://user-images.githubusercontent.com/11599942/180796105-216a5088-a11d-4aa4-8eb9-d81abea4f493.mov

marcoroth:
As pointed out by @ocarreterom the `useIntersection` mixin would just fire the `disappear` event once when the mixin is being used with multiple elements in the same controller.

This PR fixes this issue by moving the tracking reference from the controller instance to each element which is being tracked.

https://user-images.githubusercontent.com/6411752/183229395-bf4b37b6-16b3-465a-a62d-eee9f43f1c9d.mp4

Resolves #211

### DIFF
--- a/src/use-intersection/use-intersection.ts
+++ b/src/use-intersection/use-intersection.ts
@@ -71,11 +71,7 @@ export const useIntersection = (controller: IntersectionComposableController, op
 
   Object.assign(controller, {
     get isVisible() {
-      if (controller.intersectionElements.length === 1) {
-        return controller.intersectionElements[0].hasAttribute('isVisible')
-      } else {
-        return controller.intersectionElements.every(element => element.hasAttribute('isVisible'))
-      }
+      return controller.intersectionElements.every(element => element.hasAttribute('isVisible'))
     },
     get noneVisible() {
       return controller.intersectionElements.filter(element => element.hasAttribute('isVisible')).length === 0
--- a/src/use-intersection/use-intersection.ts
+++ b/src/use-intersection/use-intersection.ts
@@ -20,13 +20,13 @@ export const useIntersection = (controller: IntersectionComposableController, op
     const [entry] = entries
     if (entry.isIntersecting) {
       dispatchAppear(entry)
-    } else if (targetElement.hasAttribute("isVisible")) {
+    } else if (targetElement.hasAttribute('isVisible')) {
       dispatchDisappear(entry)
     }
   }
 
   const dispatchAppear = (entry: IntersectionObserverEntry) => {
-    targetElement.setAttribute("isVisible", "true")
+    targetElement.setAttribute('isVisible', 'true')
     method(controller, 'appear').call(controller, entry)
 
     // emit a custom "appear" event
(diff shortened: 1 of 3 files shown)


src/use-intersection/use-intersection.ts
```diff
@@ -66,7 +66,6 @@ export const useIntersection = (controller: IntersectionComposableController, op
   }
 
   Object.assign(controller, {
-    isVisible: false,
```
ocarreterom, line 69: Should we keep this attribute with something like this?

```js
Object.assign(controller, {
  get isVisible () {
    if (controller.element === targetElement) {
      return controller.element.hasAttribute('isVisible')
    }
  }
})
```

marcoroth: Hmm, good idea. But the problem is kinda the same: which element do you ask if it's visible if you have more than one element per controller?

ocarreterom: That's the reason of the `if`. Only return `true/false` if the target is the controller, `undefined` otherwise.

Also the getter could be override only if the controller is the target.

```js
if (controller.element === targetElement) {
  Object.assign(controller, {
    get isVisible () {
      return controller.element.hasAttribute('isVisible')
    }
  })
}
```

marcoroth: Oh, I see - my bad. Yeah that could work.

marcoroth: Another idea would to have a "at least one is visible" and/or a "all are visible" helper

ocarreterom: > Another idea would to have a "at least one is visible" and/or a "all are visible" helper

That will be perfect for my usecase since I'm building a responsive horizontal menu a la GitHub 😆

marcoroth: Just added that, how do you feel about it?

ocarreterom: Yes, that seems too specific for me. I can do it myself in a simple way.

```js
import { Controller } from '@hotwired/stimulus'
import { useIntersection } from 'stimulus-use'

  static targets = ['item']

  get allVisible () {
    return this.itemTargets.every(item => item.hasAttribute('isVisible'))
  }

  connect () {
    this.itemTargets.forEach(item => useIntersection(this, { element: item }))
  }

  appear () {
    this.update()
  }

  disappear () {
    this.update()
  }

  update () {
    if (this.allVisible) {
      // ...
    }
  }
}
```

marcoroth: I'm happy to keep the helper functions in there, I just was just referring to callbacks and that it would be too much too add them as individual ones. 

I'll update the docs and then I think we are good to go!

marcoroth:
@all-contributors add @ocarreterom for code review

PR #221, opened 2022-08-11
Fix language formatting in Github

ollietb:


marcoroth:
Hey @ollietb, thanks for opening this!

Technically those markers are to highlight the relevant code-lines in the docs. Though breaking the code formatting on GitHub is not nice.

Screenshot 2022-08-11 at 17 37 31

marcoroth (approved):
I just figured, that the highlights are not even respected in the current docs. As long as we are not using `vitepress` (via https://github.com/stimulus-use/stimulus-use/pull/215) it doesn't make sense to have those markers in the code blocks.

PR #226, opened 2022-10-15
fix: use-transition default options assignment

mjc-gh:
I think I found a small bug in this mixin where the `defaultOptions` object was getting overridden with values from callers due how `Object.assign` was being used.

I think the fix is to just assign the defaults and then the caller options to a new object.


marcoroth (approved):
Good catch, thank you @mjc-gh!

marcoroth:
@all-contributors add @mjc-gh for bug

PR #230, opened 2022-10-24
Mark `useDispatch` as deprecated

### LINKED ISSUE #206: `useDispatch`: deprecate mixin
Since Stimulus now got a `dispatch()` function built-in (see: https://github.com/hotwired/stimulus/pull/302) we should think about deprecating (and later removing) the `useDispatch()` mixin from Stimulus Use. 

While it's still handy to have it included, so older apps can also use it, we should go through a deprecation cycle and start printing deprecation warnings to people still using it and let them know how they can migrate it.

Follow-up of #174

leon-vogt:
closes #206 

Is the deprecation message good enough?
```
useDispatch() is a deprecated mixin from Stimulus-Use. Please use the built-in dispatch() function from Stimulus: https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events
```

Two questions:
1. Should other docs also be updated?
- docs/application-controller.md
- docs/application.md
- docs/debug.md
2. Should a "deprecation" file be created, to track the deprecation cycle?

### DIFF
--- a/docs/use-dispatch.md
+++ b/docs/use-dispatch.md
@@ -2,6 +2,32 @@
 
 Adds a `dispatch` helper function to emit custom events. Useful to communicate between different controllers.
 
+
+!> **Deprecated**: `useDispatch()` is deprecated. Please use the built-in `this.dispatch()` function from Stimulus: https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events
+
+## Migration guide
+
+Because the `dispatch()` function from Stimulus is very similar, the migration process to the Stimulus version of `dispatch()` should be fairly simple:
+- remove the `useDispatch` import
+- remove the  `useDispatch` initializer
+- wrap your payload in a `detail` object
+
+```js
+  import { Controller } from '@hotwired/stimulus'
+- import { useDispatch } from 'stimulus-use'
+
+  export default class extends Controller {
+    connect() {
+-     useDispatch(this)
+    }
+
+    add() {
+-     this.dispatch('add', { quantity: 1 })
++     this.dispatch('add', { detail: { quantity: 1 } })
+    }
+  }
+```
+
 ## Reference
 
 #### Mixin
--- a/src/stimulus-use.ts
+++ b/src/stimulus-use.ts
@@ -55,6 +55,10 @@ export class StimulusUse {
     this.logger.groupEnd()
   }
 
+  warn = (message: string): void => {
+    this.logger.warn(`%c${this.controller.identifier} %c${message}`, 'color: #3B82F6')
+  }
+
   dispatch = (eventName: string, details: any = {}) => {
     if (this.dispatchEvent) {
       const { event, ...eventDetails } = details
--- a/src/use-dispatch/use-dispatch.ts
+++ b/src/use-dispatch/use-dispatch.ts
@@ -50,7 +50,7 @@ export class UseDispatch extends StimulusUse {
     targetElement.dispatchEvent(event)
 
     warn(
-      'useDispatch() is a deprecated mixin from Stimulus-Use. Please use the built-in this.dispatch() function from Stimulus. You can find more information at: https://stimulus-use.github.io/stimulus-use/#/use-dispatch'
+      '`useDispatch()` is deprecated. Please use the built-in `this.dispatch()` function from Stimulus. You can find more information on how to upgrade at: https://stimulus-use.github.io/stimulus-use/#/use-dispatch'
     )
     log('dispatch', { eventName: eventNameWithPrefix, detail, bubbles, cancelable })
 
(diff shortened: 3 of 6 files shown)


docs/use-dispatch.md
```diff
@@ -2,6 +2,8 @@
 
 Adds a `dispatch` helper function to emit custom events. Useful to communicate between different controllers.
 
+**Deprecated** : useDispatch() is a deprecated mixin from Stimulus-Use. Please use the built-in dispatch() function from Stimulus: https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events
```
marcoroth, line 5:
```suggestion
**Deprecated**: `useDispatch()` is deprecated. Please use the built-in `this.dispatch()` function from Stimulus: https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events
```

src/stimulus-use.ts
```diff
@@ -55,6 +55,10 @@ export class StimulusUse {
     this.logger.groupEnd()
   }
 
+  warn = (functionName: string, args: any): void => {
```
marcoroth, line 58: I think you can leave out the function name here and remove it from the `warn` function signature. It should anyway be visible from the backtrace itself.

src/use-dispatch/use-dispatch.ts
```diff
@@ -49,6 +49,7 @@ export class UseDispatch extends StimulusUse {
     // dispatch the event from the given element or by default from the root element of the controller
     targetElement.dispatchEvent(event)
 
+    warn('dispatch', 'useDispatch() is a deprecated mixin from Stimulus-Use. Please use the built-in dispatch() function from Stimulus: https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events');
```
marcoroth, line 52: This doesn't seem to work. 

Ironically if you have this controller here:

```js
import { Controller } from '@hotwired/stimulus'
import { useDispatch } from 'stimulus-use'

  connect() {
    useDispatch(this)
  }

  add() {
    this.dispatch('add', { quantity: 1 })
  }
}
```
It's already using the `this.dispatch()` function from Stimulus itself. So this code in our `dispatch` function is actually never called again. 

Which is why we should move the deprecation warning into our `useDispatch()` function itself.

marcoroth (changes requested):
Hey @leon-vogt, thanks for taking this on and opening this pull requests! 🤗 

> 1. Should other docs also be updated?

Yes, I think that would make sense!

> 2. Should a "deprecation" file be created, to track the deprecation cycle?

I don't think we need that (yet). It should be fine to have the warning in there. And once we remove the function itself we can also get rid of the deprecation warning with it.

I left some comments, but this looks good to me otherwise! 🙌🏼

leon-vogt:
@marcoroth thanks for taking the time and review the code!
I will update the other docs as well and adopt your suggestions

src/use-dispatch/use-dispatch.ts
leon-vogt: Hm, that's odd...
When I run your controller example with two additional console.log outputs:

```js
import { Controller } from '@hotwired/stimulus'
import { useDispatch } from 'stimulus-use'

  connect() {
    console.log("ITEM Controller connected");
    useDispatch(this)
  }

  add() {
    console.log("ITEM Controller#add was called");
    this.dispatch('add', { quantity: 1 })
  }
}
```

I get following Output in the Console:

image


---

The Reason why I didn't put the deprecation warning into the `useDispatch()` function itself is, it would print the warning on every request, even though I don't use or import the `useDispatch()` function.
That's why I put the deprecation warning in the `dispatch()` call.

Do you thinks that's a bad idea?

leon-vogt:
The deprecation warning is now in the documentation visible and looks like this:
image

I removed the `useDispatch` example from `docs/debug.md` (To propose not outdated things)
And added deprecation warnings in the following files:
- docs/application-controller.md
- docs/application.md
- docs/use-dispatch.md

--- 

The one thing I wasn't sure about, was the [cart example](https://stimulus-use.github.io/stimulus-use/#/application-controller?id=example-building-a-cart-counter-with-the-dispatch-helper)...
Maybe an example with another mixin like `isPreview` would make more sense now.

marcoroth:
Thank you, this looks very good now!

> The one thing I wasn't sure about, was the [cart example](https://stimulus-use.github.io/stimulus-use/#/application-controller?id=example-building-a-cart-counter-with-the-dispatch-helper)... Maybe an example with another mixin like `isPreview` would make more sense now.

You are right. Either we remove it, or we just keep it in there to actually have a demo for the `dispatch` function in general. But if you ask me we can just keep it there for now.

marcoroth (approved):
Awesome, thank you @leon-vogt for pulling this one off! 🎉

PR #244, opened 2022-11-24
`useHotKeys`: Allow unbinding hotkeys within a scope if provided

### LINKED ISSUE #177: useHotkeys does unbind all handlers attached to a key instead of just the one that the controller is disconnecting
For example, let's suppose we have to forms on the same page, each one attached to a controller that registers the same hotkey for both, when one of those forms is submitted the controller disconnects and unbinds the `key`, but it is unbinding both handlers for that given key instead of just that one of the form that was submitted.

Ex:
```html
<form data-controller="example" id="new_name">
  <label for="name">Name:</label><br>
  <input type="text" id="name" name="name"><br>
  <input type="submit" name="commit" value="Submit" id="submit_new_name">
</form>
<!-- some other HTML here -->
<form data-controller="example" id="new_message">
  <label for="message">Message:</label><br>
  <input type="text" id="message" name="message"><br>
  <input type="submit" name="commit" value="Submit" id: "submit_new_message">
</form>
```

```js
import { Controller } from "@hotwired/stimulus"
import { useHotkeys } from "stimulus-use"

// Connects to data-controller="example"
  connect() {
    useHotkeys(this, {
      hotkeys: {
        "ctrl+enter": {
          handler: this.submitForm,
          options: {
            element: this.element
          }
        },
        "cmd+enter": {
          handler: this.submitForm,
          options: {
            element: this.element
          }
        }
      },
      filter: this.filterForInput
    })
  }

  submitForm (evt) {
    const formToSubmit = evt.target.closest("form")
    const submitButton = formToSubmit.querySelector(`#submit_${formToSubmit.id}`)
    if (submitButton.disabled) return;

    submitButton.focus();
    submitButton.click();
    submitButton.disabled = true;
  }

  filterForInput(evt) {
    return evt.target.tagName === "INPUT"
  }
}
```

In this example above each form can be submitted when pressing `Ctrl+Enter`, if I for example submit one of those first, I can go to the next and provide a value to the input and press `Ctrl+Enter` and it won't submit, since the first submit did unbind the key for all handlers and not only for the handlers attached to that form. 

From what I can see when Hotkeys.js unbinds a key it accepts as args the scope and the handler https://github.com/jaywcjlove/hotkeys/blob/master/src/index.js#L112 and here https://github.com/jaywcjlove/hotkeys/blob/master/src/index.js#L142 it checks if the method of a given handler is the same as the one that is being unbound, if `use-hotkeys` pass the handler to the `unbind` method then It wouldn't unbind handlers that it shouldn't. 

I try it here and had to make the following changes to make it work as expected. I'm also not sure why in the `bind` method the handler is passed down as `(e: KeyboardEvent) => handler(e, e as unknown as HotkeysEvent)` instead of just `handler` (this would be a new function and never would match the handler passed down in the unbind method)

https://github.com/stimulus-use/stimulus-use/blob/main/src/use-hotkeys/use-hotkeys.ts#L42
```diff
bind = () => {
    for (const [hotkey, definition] of Object.entries(this.hotkeysOptions.hotkeys as any)) {
      const handler = (definition as HotkeyDefinition).handler.bind(this.controller)
-       hotkeys(hotkey, (definition as HotkeyDefinition).options, (e: KeyboardEvent) =>
-          handler(e, e as unknown as HotkeysEvent)
-        )
+      hotkeys(hotkey, (definition as HotkeyDefinition).options, handler)
    }
  }
```

https://github.com/stimulus-use/stimulus-use/blob/main/src/use-hotkeys/use-hotkeys.ts#L51
```diff
unbind = () => {
-    for (const hotkey in this.hotkeysOptions.hotkeys as any) {
-      hotkeys.unbind(hotkey)
+   for (const [hotkey, definition] of Object.entries(this.hotkeysOptions.hotkeys as any)) {
+     const handler = (definition as HotkeyDefinition).handler.bind(this.controller)
+     hotkeys.unbind(hotkey, (definition as HotkeyDefinition).scope, handler)
    }
  }
```

tkoenig:
This partially fixes #177 , as it at least now allows to unbind the shortcuts within a scope if one has been provided.

## Example 

```hotkey_controller.js
import { Controller } from "@hotwired/stimulus";
import { useHotkeys, useThrottle } from "stimulus-use";

  static values = { key: String, scope: { type: String, default: "app" } };
  static throttles = ["fire"];
  connect() {
    useThrottle(this, { wait: 2000 });
    hotkeys = {};
    hotkeys[this.keyValue] = { handler: this.fire, options: { scope: this.scopeValue } };
    useHotkeys(this, { hotkeys: hotkeys });
  }

  fire() {
    this.element.click();
  }
}

```

```html

<button data-action="dialog#close" data-controller="hotkey" data-hotkey-scope-value="preview" data-hotkey-key-value="Esc" type="button">
  Close
</button>


<button data-action="selection#empty" data-controller="hotkey" data-hotkey-scope-value="selection" data-hotkey-key-value="Esc" type="button">
  Empty selection
</button>
```

So when Closing the dialog, the `Esc` key gets unbound correctly, while the Empty selection hotkey will still work.

### DIFF
--- a/src/use-hotkeys/use-hotkeys.ts
+++ b/src/use-hotkeys/use-hotkeys.ts
@@ -49,8 +49,10 @@ export class UseHotkeys extends StimulusUse {
   }
 
   unbind = () => {
-    for (const hotkey in this.hotkeysOptions.hotkeys as any) {
-      hotkeys.unbind(hotkey)
+    for (const [hotkey, definition] of Object.entries(this.hotkeysOptions.hotkeys as any)) {
+      const options = (definition as HotkeyDefinition).options
+      const scope = options.scope ? options.scope : ''
+      hotkeys.unbind(hotkey, scope)
     }
   }
 


tkoenig:
Actually it was easier using `hotkeys.js` directly and since stimulus will get support for keyboard events (https://github.com/hotwired/stimulus/pull/442) there is probably no more need for that.

PR #262, opened 2023-01-14
Extract `useHotkeys()` and remove `hotkeys-js` as dependency

### LINKED ISSUE #237: Possible to exclude hotkeys from generated JS?
I'm not using they hotkeys mixin, but `hotkeys-js` still ends up being included in my application.js

I think it's because when I use a mixin like`useIntersection` it will include `stimulus-use/index.js` which immediately imports `hotkeys-js`.

I'm not familar with Typescript, so I can't point to the source code that causes this. But I see the compiled `node_modules/stimulus-use/dist/index.js` which shows this:

```javascript
/*
Stimulus-Use 0.50.0
*/
import hotkeys from 'hotkeys-js';
```

Would it make sense to make this import part of the hotkeys mixin somehow? So it only gets imported when the hotkeys mixin is used.

marcoroth:
This pull request moves the `useHotkeys()` behavior to separate file, so that the main bundle doesn't include `hotkeys-js` as a NPM dependency.

If you are using `useHotkeys()` in your application and you are upgrading to the next version (likely `v0.52.0`) you have to change the import for the `useHotkeys()` behavior.

```diff
-import { useHotkeys } from "stimulus-use"
+import { useHotkeys } from "stimulus-use/hotkeys"
```

Also, please add `hotkeys-js` as a npm dependency to your application if you haven't added it yet, it's not being removed as a runtime  dependency.

The existing import (`import { useHotkeys } from "stimulus-use"`) will throw an error in the new version, so that people know how to upgrade (if we don't throw the error here it would throw an error down the stack, because the `import "hotkeys-js"` would fail):

Screenshot 2023-01-14 at 18 31 54

Resolves #237

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,6 +7,7 @@
 tsconfig.json
 webpack.config.js
 *.log
+*~
 karma.conf.js
 yarn-error.log
 /.git
--- a/docs/use-hotkeys.md
+++ b/docs/use-hotkeys.md
@@ -1,16 +1,29 @@
 # useHotkeys
 
-Adds one new behavior to your Stimulus controller : `useHotkeys`
+The `useHotkeys` behavior can be used to register hotkeys using the [hotkeys-js](https://wangchujiang.com/hotkeys/) library. To use it, do the following:
 
-This behavior can be used to register hotkeys using the [hotkeys-js](https://wangchujiang.com/hotkeys/) library. To use it, do the following:
-
-1. Install the hotkeys-js peer dependency:
+1. Install the `hotkeys-js` peer dependency:
 
 ```bash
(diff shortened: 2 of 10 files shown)


rainerborene:
It seems that `hotkeys-js` is still included on `dist/index.js` file.
```js
/*
Stimulus-Use 0.51.3
*/
import hotkeys from 'hotkeys-js';
``

marcoroth:
Hey @rainerborene! Yeah, this pull request isn't released yet. I'm going to release `0.52.0` soon-ish!

marcoroth:
I just released `0.52.0`, let me know if you still see something unexpected!

PR #271, opened 2023-01-24
Add dev-builds

marcoroth:
This pull request adds a GitHub Actions workflow for creating stimulus-use dev-builds at https://github.com/stimulus-use/dev-builds


adrienpoly:
this is awesome @marcoroth lets try it

marcoroth:
It just needs the `STIMULUSUSE_DEV_BUILDS_TOKEN` secret to be set. GitHub now also supports org tokens which should be ideal for this.

Issue #370, opened 2023-07-07
`useIntersection`: unobserve after triggering `appear` callback


Eldzej:
Is it possible to somehow call the `appear` event added by `useIntersection` only once? Or manually call `unobserve` on the created `IntersectionObserver`?

In my usecase, I am initializing a library when element enters the viewport, but with the default options, the code gets invoked every time when element enters the viewport, which is rather impractical.

For the moment, I am managing it by checking a custom property on the controller, but wondering if there isn't a "cleaner" way

```
connect() {
        this.isLibraryLoaded = false
        useIntersection(this)
    }

    appear() {
        if (!this.isLibraryLoaded) {
            this.isLibraryLoaded = true
            this.loadLibrary()
        }
    }

    loadLibrary() {
        /* do stuff */
    }
```

Sorry if this is stupid question or answered somewhere, I could not find it...

marcoroth:
Hey @Eldzej, currently there is no built-in option to just call it once. But you can call `unobserve` manually by obtaining a reference to the function from the `useIntersection()` function call. It returns an array with two functions, the first one is `observe` and the second one `unobserve`. So you could do something like:
```js
export default class extends Controller { 
  connect() {
    const [_observe, unobserve] = useIntersection(this)
    this.unobserve = unobserve
  }
  
  appear(entry) {  
    console.log("appear", entry)
    this.unobserve()
  }
}
```
Here's a working example: https://codepen.io/marcoroth/pen/YzRrrRr?editors=1010

But the brings up the question. I feel like we could also expose the actual `IntersectionObserver` instance by either assigning it to a property on the controller or by passing in the reference to the `appear` and `disappear` callback as the second argument.

So that you could do something like:
```js
export default class extends Controller { 
  connect() {
    useIntersection(this)
  }
  
  appear(entry, observer) {  
    console.log("appear", entry)
    observer.unobserve(entry.target)
  }
}
```

I hope this helps

Eldzej:
Yeah, that is something I have been trying, thanks.

I have been trying to get the reference incorrectly (looking at code, I thought it should be object, rather than array), and I kept getting undefined as a result

I think exposing the observer instance as part of the controller (similar to how the helper functions are) itself might be good? That is how I have been using IntersectionObserver before stumbling upon this repo.

```
this.observer = new IntersectionObserver(entries => {
    if (entries[0].isIntersecting && this.inView !== true) {
        this.inView = entries[0].isIntersecting
[...]

PR #371, opened 2023-07-07
Pass `IntersectionObserver` instance to `useIntersection` callbacks

marcoroth:
This pull request allows the callbacks from the `useIntersection` behavior to access the `IntersectionObserver` as the second argument in the callback.

This enables things like this:
```js
  connect() {
    useIntersection(this)
  }
  
  appear(entry, observer) {  
    observer.unobserve(entry.target)
  }
}
```

This allows you to call all available methods directly on the `observer` itself to implement the logic according to your needs. In this case it would allow to only trigger the `appear` callback once.

This enables a more elegant solution for https://github.com/stimulus-use/stimulus-use/issues/370

### DIFF
--- a/docs/use-intersection.md
+++ b/docs/use-intersection.md
@@ -1,6 +1,6 @@
 # useIntersection
 
-Adds 2 new behaviors to your Stimulus controller : `appear` and `disappear`.
+Adds 2 new behaviors to your Stimulus controller: `appear` and `disappear`.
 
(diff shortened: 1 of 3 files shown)


adrienpoly (approved):
Thanks @marcoroth this is a very nice solution and an improved documentation. 
I have been using the manual version a few time but passing the observer to the callback is a great alternative

PR #462, opened 2023-12-01
fix: convert html syntax to markdown syntax in readme

omarluq:
Converting to markdown syntax removes the underline you can see in the attached screenshot! it's a small detail but I find it really annoying! 
before:
Screenshot 2023-11-30 at 11 21 20 PM

after:
Screenshot 2023-11-30 at 11 29 38 PM


marcoroth (approved):
Thanks @omarluq! 🚀

PR #471, opened 2023-12-24
`useTargetMutation`: Add deprecation mention for `[target]TargetAdded` and `[target]TargetRemoved`

ker0x:
This PR adds a depreciation mention for `[target]TargetAdded` and `[target]TargetRemoved` as their behavior appears to be identical to `[target]TargetConnected` and `[target]TargetDisconnected`.

Feel free to close if it's not 😉 !

### DIFF
--- a/docs/use-target-mutation.md
+++ b/docs/use-target-mutation.md
@@ -1,6 +1,8 @@
 # useTargetMutation
 
-`useTargetMutation` tracks when targets are added and removed to your controller.
+`useTargetMutation` tracks when targets are added, updated and removed to your controller.
+
+!> **Deprecated**: `[target]TargetAdded` and `[target]TargetRemoved` are deprecated. Please use the built-in `[target]TargetConnected()` and `[target]TargetDisconnected()` functions from Stimulus: https://stimulus.hotwired.dev/reference/targets#connected-and-disconnected-callbacks
 
 It adds three new behaviors to your Stimulus controller for each target you specify:
 
--- a/src/use-target-mutation/use-target-mutation.ts
+++ b/src/use-target-mutation/use-target-mutation.ts
@@ -159,12 +159,18 @@ export class UseTargetMutation extends StimulusUse {
   private targetAdded(name: string, node: Node, trigger: string) {
     let targetCallback = `${name}TargetAdded`
     this.controller[targetCallback] && method(this.controller, targetCallback).call(this.controller, node)
+    this.warn(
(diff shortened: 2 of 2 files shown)


marcoroth:
Hey @ker0x, thanks for this, good catch!

Would you mind adding deprecation logs to the actual code as well?

https://github.com/stimulus-use/stimulus-use/blob/029624d7f1423e67ffa63ca1209a9c2afeda2b3e/src/use-target-mutation/use-target-mutation.ts

ker0x:
> Would you mind adding deprecation logs to the actual code as well?

@marcoroth Done 😉 !

marcoroth:
Awesome thanks! I have one more thing and then I think we can merge this.

Can we add another deprecation warning to the `useTargetMutation` method itself so you can already see it as soon as the `useTargetMutation` is called in the controller?

https://github.com/stimulus-use/stimulus-use/blob/efa70d81b89ba44e287ce36fa66c0eeb8850414a/src/use-target-mutation/use-target-mutation.ts#L212

ker0x:
@marcoroth I've added it, but I'm not sure about it, because it will throw a warning even if you use the [`target]TargetChanged` function, right?

marcoroth:
Oh, you are right. This almost makes me want to implement the changed target callback in upstream Stimulus.

ker0x:
@marcoroth That would be great! There was a previous discussions and attempt to introduce it in Stimulus: hotwired/stimulus#445 and hotwired/stimulus#460

Regardint the PR, should I remove the warning on `useTargetMutation` ?

marcoroth (approved):
Thank you @ker0x! 🙏🏼

marcoroth:
@all-contributors add @ker0x for code

ker0x:
@marcoroth thank you 🙏 !

brendon:
> @marcoroth I've added it, but I'm not sure about it, because it will throw a warning even if you use the [`target]TargetChanged` function, right?

I don't think this was ever removed again per: https://github.com/stimulus-use/stimulus-use/pull/471#issuecomment-1886706854

I'm getting the deprecation warning in the console regardless of whether I fix the problem or not.

brendon:
> > @marcoroth I've added it, but I'm not sure about it, because it will throw a warning even if you use the [`target]TargetChanged` function, right?
> 
> I don't think this was ever removed again per: [#471 (comment)](https://github.com/stimulus-use/stimulus-use/pull/471#issuecomment-1886706854)
> 
> I'm getting the deprecation warning in the console regardless of whether I fix the problem or not.

Oh, disregard this. I misread it as a method name change deprecation rather than indicating that the functionality now exists in stimulus itself.

PR #478, opened 2024-01-03
Add `useCookie` mixin

divagueame:
This PR adds use-cookie to stimulus-use. I tried to keep it as simple as possible and to provide the most basic functionality in regards of could be needed to easily work with cookies on the client.

I added some tests with the core functionality and added an example on the playground. I tried to test if cookies were expiring as expected but I could not get the Sinon time fakers working properly, so eventually I gave up and decided to put it out there. Maybe someone has a better idea of how get properly tested.

Glad to hear any feedback.

### DIFF
--- a/docs/use-cookie.md
+++ b/docs/use-cookie.md
@@ -1,20 +1,23 @@
 # useCookie
 
-Adds getters to easily read cookies and provides a simple API to set and clear cookies.
+Adds getters and setters to easily read cookies and provides a simple API to set and clear cookies.
 
-Getters are then automatically defined by the mixin.
-- Getter name is a camelized conversion of the cookie name with the `Cookie` suffix. `user_id`  becomes `userIdCookie`
+Simply list at the top of your controller the list of cookie names you need (static cookieNames = [])
+
+Accessors are then automatically defined by the mixin.
+- Getter/Setter name is a camelized conversion of the cookie name with the optional `Cookie` suffix. `user_id`  becomes `userIdCookie`
 
 **Options:**
 
 | Option| Description | Default value |
 |-----------------------|-------------|---------------------|
 | `expires` | Expire date of a cookie in days. Can be overriden on the options object on a per cookie basis |30|
+| `suffix` | prepend or not `Cookie` to the accessor's name. Default is true to remain consistent with `Value` and `Class` API |true|
 
 
 **Example**
 ```js
-useCookie(this, { expires: 30 }) // Same as useCookie(this)
+useCookie(this, { expires: 30, suffix: true }) // Same as useCookie(this)
 ```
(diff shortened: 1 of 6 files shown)


docs/use-cookie.md
```diff
@@ -0,0 +1,81 @@
...
+    // Or pass an object to change the expires value (default 30 days):
+    // useCookie(this, { expires: 30 })
+
+    // Getters are initialized for all current cookies.
+    // For a cookie with name 'colorScheme' and value 'dark'
+    this.colorSchemeCookie // dark
+
+    // You can save new cookies with setCookie
+    // By passing two strings as name-value
+    this.setCookie("dessert", "cake")
```
marcoroth, line 41: I'm wondering if it would be more natural from an API point of view, and also for symmetry, to use a more Stimulus-like way to set cookies. Something like:
```suggestion
    // You can save new cookies with the [name]Cookie setter
    this.desertCookie = "cake"
```

Or was there a reason against this?

docs/use-cookie.md
```diff
@@ -0,0 +1,81 @@
...
+export default class extends Controller {
+
+  connect() {
+    useCookie(this)
+    // Or pass an object to change the expires value (default 30 days):
+    // useCookie(this, { expires: 30 })
+
+    // Getters are initialized for all current cookies.
+    // For a cookie with name 'colorScheme' and value 'dark'
+    this.colorSchemeCookie // dark
```
marcoroth, line 37: I'm not sure if this is the best behavior. I can technically see how this is useful, but this somehow goes against the other Stimulus/Stimulus use patterns.

I think it would make sense if you have to explicit tell the mixin the cookie names you want to use. This makes it less magical and easier to follow the flow, especially if other people on the team want to understand the controller. But also for us, so we can appropriately generate the right getters and setters for all the defined cookies.

marcoroth (commented):
Thank you so much for working on this @divagueame.

This looks good overall, but I think there are some nuances that I personally would implement differently to be closer to the other Stimulus/Stimulus-use patterns. Take a look and let me know what you think. Thank you!

divagueame:
hey, thanks for the feedback. Both issues you commented on are kind of related. 

First, I implemented the API  like you suggested    ` this.desertCookie = "cake"` instead of `this.setCookie("dessert", "cake")` 
That would be very similar to useMeta, declaring the cookies that will be used and creating the getters/setters from those. 

I thought it would be useful to skip that step an do it magically : ) However,  I can understand the benefits  of keeping it consistent and explicit, so I can rewrite like that.

divagueame:
Hi, I refactor the PR to follow a more similar API to the rest of the library. Let me know what you think.

PR #482, opened 2024-01-11
Fix broken links to hotkeys-js

BarnabeD:
I noticed that the links to the Hotkeys-js library were broken, so I fixed them.
It might be interesting to link directly to the Github repository rather than the official site? what do you think? 
I'll make the change if you prefer.


marcoroth (approved):
Thank you @BarnabeD!

PR #559, opened 2024-12-18
feat(useDebounce): Add `leading` and `trailing` options

Kerrick:
This change introduces a new `leading` boolean option to the useDebounce utility. It defaults to `false`. When enabled, the function will run immediately, on the ["leading" edge](https://css-tricks.com/debouncing-throttling-explained-examples/#aa-leading-edge-or-immediate) of the repeated calls. It does not prevent the function from also running eventually, on the "trailing" edge of the repeated calls.

It also introduces a `trailing` boolean option, which defaults to `true` so the existing behavior is unchanged. When set to `false`, the function will _not_ run on the trailing edge — so `{ leading: true, trailing: false }` gives you a leading-edge-only debounce. (Setting both to `false` means the function never runs.)

My changes include documentation, playground examples, and tests. The default behavior remains unchanged for existing uses of useDebounce.

## Motivation

I'm developing an application where I have a function that I want to run on `input`. I need `useDebounce` because a fast typer can bog down the thread with lots of calls to the function. But a consistent typist might not stop typing for a while. For those users, the experience is worse because the function that should've already run waits for them to finish. Calling the function immediately improves the user experience. I am okay if the function _also_ runs after they've stopped typing, as it's idempotent.

## Alternatives

The `trailing` option (now included) covers the leading-edge-only case via `{ leading: true, trailing: false }`. For more advanced needs like `maxWait` or `cancel`/`flush`, it may be worth using `lodash/debounce`.


marcoroth (approved):
Thanks @Kerrick!

I ended up adding the `trailing` option too for completeness sake.

PR #560, opened 2024-12-19
docs(getting_started): Add importmap-rails section (Fixes #529)

### LINKED ISSUE #529: Docs request: how to get started with import maps
I got a little hung up trying to start using stimulus-use, because my app uses Rails 7 importmaps (no yarn, no webpacker, etc.) and its only mention in the docs is [stimulus-use/hotkeys](https://stimulus-use.github.io/stimulus-use/#/use-hotkeys), which suggested to me that every plugin is published separately (which they apparently are not)

For me, here's how setup looked. Would be nice if this was in the install docs, but will leave this to feed Google for anyone else looking for it:

```sh
bin/importmap pin stimulus-use
```

Which (in a default confIg) will append this to `importmap.rb`:

```ruby
pin "stimulus-use" # @0.52.2
```

It will also vendor stimulus-use in `vendor/javascript/stimulus-use.js`

**BUT**

It will also, unfortunately, redundantly vendor stimulus itself to `vendor/javascript/@hotwired--stimulus.js`, which should be removed if using the `stimulus-rails` gem (which most people will be). 

So all together, my suggested setup for importmap users is:

```
bin/importmap pin stimulus-use
rm vendor/javascript/@hotwired--stimulus.js
```

And then edit config/importmap.rb to revert it back to:

```diff
- pin "@hotwired/stimulus", to: "@hotwired--stimulus.js" # @3.2.2
+ pin "@hotwired/stimulus", to: "stimulus.min.js"
```

After this, the thing I wanted worked:

```js
import { useIntersection } from 'stimulus-use'
```

Kerrick:
Thanks @searls!


indigotechtutorials:
I was gonna add my own PR because I saw this section missing. Wonder why this never got merged in.

marcoroth (approved):
Thanks @Kerrick! 🙏🏼

marcoroth:
@all-contributors add @Kerrick for docs

PR #569, opened 2025-03-11
Extra type safety check in use-click-outside

adamyonk:
I'm seeing cases where an error arises from assuming `event.target` is always a `Node`:

```
TypeError: Argument 1 ('other') to Node.contains must be an instance of Node
```

I'd like to propose adding an extra check here so it doesn't fail if it is `Window` for instance.

### DIFF
--- a/src/use-click-outside/use-click-outside.ts
+++ b/src/use-click-outside/use-click-outside.ts
@@ -24,7 +24,10 @@ export const useClickOutside = (composableController: Controller, options: Click
   const onEvent = (event: Event) => {
     const targetElement: Element = options?.element || controller.element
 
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Good catch, thanks for the contribution @adamyonk! 🙏🏼

marcoroth:
@all-contributors add @adamyonk for code

adamyonk:
Is there a planned release cycle, or when could we expect to see this in the main release?

adamyonk:
Oh, I just saw the `dev-builds` repo. If the workflow could be fixed I think we could just use that for the time being!

PR #584, opened 2026-06-20
fix: forward ElementType generic to Controller

### LINKED ISSUE #583: `ApplicationController` does not support a type parameter
`ApplicationController` extends Stimulus `Controller` without forwarding the `ElementType` generic parameter, making it harder to for downstream projects to type their code.

myabc:
## What

`ApplicationController` extends Stimulus `Controller` without forwarding the `ElementType` generic parameter, so `this.element` is always typed as `Element`. Subclasses cannot narrow it (e.g. to `HTMLFormElement`) without a cast.

This adds the generic parameter and forwards it to `Controller<ElementType>`:

```ts
export class ApplicationController<ElementType extends Element = Element> extends Controller<ElementType> {
```

## Usage

```ts
class FormController extends ApplicationController<HTMLFormElement> {
  submit() {
    this.element.requestSubmit() // typed, no cast
  }
}
```

## Compatibility

The default `= Element` preserves backward compatibility — existing `extends ApplicationController` is unaffected.

## Verification

- `tsc` typecheck passes; emitted `.d.ts` declares the generic.
- Full test suite passes (123/123).

Closes #583

### DIFF
--- a/src/use-application/application-controller.ts
+++ b/src/use-application/application-controller.ts
@@ -2,7 +2,7 @@ import { Controller, Context } from '@hotwired/stimulus'
 import { useApplication } from './use-application'
 import { DispatchOptions } from '../use-dispatch'
 
-export class ApplicationController extends Controller {
+export class ApplicationController<ElementType extends Element = Element> extends Controller<ElementType> {
   options?: DispatchOptions
   readonly isPreview: boolean = false
   readonly isConnected: boolean = false


marcoroth (approved):
Great catch, thanks @myabc! 🙏🏼

marcoroth:
@all-contributors add @myabc for code

marcoroth:
This should be released as part of [`v0.53.0`](https://github.com/stimulus-use/stimulus-use/releases/tag/v0.53.0) 🎉

PR #585, opened 2026-06-20
fix: silence karma TS18028 and reload errors

myabc:
## Problem

`yarn test` passes all specs but the runner prints errors and exits non-zero, masking real failures in CI:

```
ERROR [compiler.karma-typescript]: node_modules/puppeteer/lib/types.d.ts(...): error TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.
...
Chrome Headless ERROR
  Some of your tests did a full page reload!
```

Both are config-level, not test-level (an empty single test reproduces the reload error; the TS18028 noise comes from dependency typings).

## Root causes & fixes

**1. TS18028 from puppeteer typings**
`karma.conf.js` had no `karmaTypescriptConfig`, so karma-typescript walked `node_modules` and compiled dependency `.d.ts` files (puppeteer) at its default low target — puppeteer's `#private` fields then error. Scoped compilation to our source with `target: es2017` and `exclude: ['node_modules']`.

**2. "Some of your tests did a full page reload!"**
`client.clearContext: false` is only meant for interactive `karma start` debugging. Under headless `singleRun` it makes Karma flag a false-positive full page reload and exit with an error. Restored the Karma default (`true`).

## Output

| Before | After |
|--------|--------|
| Screenshot 2026-06-20 at 14 46 06| Screenshot 2026-06-20 at 14 45 35|

## Verification

- Before: `Executed 123 of 123 SUCCESS` but `ERROR` exit, TS18028 + reload noise.
- After: `Executed 123 of 123 SUCCESS`, clean output, `exit 0`.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -91,6 +91,18 @@ module.exports = function (config) {
     // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
     frameworks: ['karma-typescript', 'mocha', 'sinon-chai', 'fixture'],
 
+    // Scope karma-typescript to our source only. Without this it walks
+    // node_modules and compiles dependency typings (e.g. puppeteer) at its
+    // default low target, emitting TS18028 errors for their `#private` fields.
+    karmaTypescriptConfig: {
+      compilerOptions: {
+        target: 'es2017',
+        module: 'commonjs',
+        lib: ['dom', 'es2017']
+      },
+      exclude: ['node_modules']
+    },
+
     // list of files / patterns to load in the browser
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thank you for this fix, @myabc! 

Just as a FYI: we might migrate to Vitest soon to modernize the setup here 🙌🏼

myabc:
@marcoroth that makes sense! We just migrated opf/openproject from Karma to vitest. Let me know if you need q hand with that.

